PR #4, opened 2022-09-26
Add script to build Stratum dependencies

ffoulkes:
- Created a CMake script to build and install the Stratum dependencies.

- Removed obsolete OVS patch file.


ravi861 (approved):

setup/CMakeLists.txt
```diff
@@ -0,0 +1,222 @@
...
+
+project(dependencies VERSION 1 LANGUAGES C CXX)
+
+include(ExternalProject)
+include(CMakePrintHelpers)
+
+option(USE_SUDO       "Use sudo when installing" OFF)
+
+if(USE_SUDO)
+  set(SUDO_CMD "sudo")
```
ravi861, line 14: sudo -E
isn't this better if we are to capture any existing proxy settings ?

ffoulkes: Done

setup/CMakeLists.txt
```diff
@@ -0,0 +1,219 @@
...
+  GIT_PROGRESS    ON
+
+  SOURCE_DIR
+    ${CMAKE_SOURCE_DIR}/gtest
+  CMAKE_ARGS
+    -DCMAKE_INSTALL_PREFIX=${CMAKE_INSTALL_PREFIX}
+    -DBUILD_SHARED_LIBS=on
+  INSTALL_COMMAND
+    ${SUDO_CMD} ${CMAKE_MAKE_PROGRAM} install
+)
```
nupuruttarwar, line 199: Maybe we don't need gtest and gmock_gbl, we can add these when add support for unit tests?

ffoulkes: I included googletest and gmock-global because they were part of the existing package.

Incidentally, the mock-global package installation is defective. It doesn't add anything to the install tree!

My inclination is to:
1.  Make googletest and gmock-global conditional, under control of CMake options.
2.  Install `include/gmock-global/gmock-global.h` if gmock-gbl is enabled.

Long-term, we need to differentiate between development installations (which would normally _include_ header files and support for unit tests) and target installations (which would normally _omit_ header files and support for unit testing).

nupuruttarwar: I like the cmake option idea and ability to differentiate development and target installations. We can add this when adding support for different build types

PR #11, opened 2022-09-28
Add CMake support for DPDK build

ffoulkes:

### DIFF
--- a/stratum/CMakeLists.txt
+++ b/stratum/CMakeLists.txt
@@ -295,25 +295,25 @@ if(WITH_OPENCONFIG)
     target_link_libraries(stratum PUBLIC openconfig_proto)
 endif()
 
+find_library(LIBTDI tdi ${SDE_INSTALL_PREFIX}/lib)
+if(NOT LIBTDI)
+    message(FATAL_ERROR "Cannot find library: tdi")
+endif()
+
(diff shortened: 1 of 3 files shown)


stratum/CMakeLists.txt
```diff
@@ -295,55 +295,58 @@ if(WITH_OPENCONFIG)
...
         ${LIBTDI}
         ${LIBTDI_JSON_PARSER}
         ${LIBTARGET_SYS}
     )
-else()
-    find_library(TARGET_UTILS target_utils ${SDE_INSTALL_PREFIX}/lib)
-    if(NOT TARGET_UTILS)
-        message(FATAL_ERROR "Cannot find library: target_utils")
+elseif(DPDK_TARGET)
+    find_library(LIBDRIVER driver ${SDE_INSTALL_PREFIX}/lib)
```
ravi861, line 326: is this a recent change? Before this, DPDK had a requirement for libbf_switchd_lib or something like that?
libdriver was only introduced with cmake in p4studio,

nupuruttarwar (approved):
LGTM

stratum/CMakeLists.txt
ffoulkes: We still link against libbf_switchd_lib. It's the item after this one in the CMake file.

PR #14, opened 2022-10-04
Update gRPC dep build

ravi861:
gRPC builds ssl using the boring ssl submodule that comes with grpc. This is incompatible with cpython which is build to support TDI CLI causing a segfault. This change will let gRPC use the default ssl lib available in the system instead of building it from scratch.


nupuruttarwar (approved):
LGTM

ffoulkes (commented):
1. We may want to characterize the failure further. Google uses gRPC and Python extensively. I find it difficult to believe that BoringSSL and Python are universally incompatible.
2. I'll need to advise EPG upstream of the change. Their security folks expressly asked whether we were using OpenSSL or BoringSSL. This is a change from what I told them.

ffoulkes:
> Fair enough. We will investigate further if Python is compatible with boring SSL. Are there objections to merging this for now to unblock QA?

Unblocking QA is the higher priority.

ravi861:
https://peps.python.org/pep-0644/#boringssl
Apprarently, python does not support boringssl.

PR #15, opened 2022-10-06
Add support for Tofino-specific Stratum TDI files

ffoulkes:
- Moved files to new tdi/tofino subdirectories and renamed them.


nupuruttarwar (approved):
LGTM

PR #19, opened 2022-10-10
Modularize yang_parse_tree_paths

ffoulkes:
- Divided `yang_parse_tree_paths.cc` into multiple source files.
- Moved yang files to `stratum/hal/lib/yang`.
- Updated CMake build files.


nupuruttarwar (approved):
LGTM

PR #34, opened 2022-10-31
Add Runpath support

ffoulkes:


stratum/CMakeLists.txt
```diff
@@ -348,58 +331,110 @@ target_link_libraries(stratum PUBLIC
...
+    absl::log_severity absl::civil_time absl::civil_time absl::time_zone
+    absl::status absl::cord absl::cord_internal absl::cordz_info
+    absl::cordz_handle absl::cordz_sample_token absl::cordz_functions
+    absl::exponential_biased absl::str_format_internal absl::hash
+    absl::raw_hash_set absl::city absl::bad_optional_access
+    absl::bad_variant_access absl::low_level_hash
+)
+
+target_link_libraries(stratum PUBLIC pthread)
+
```
ffoulkes, line 351: The long-term plan should be to create CMake package files for the DPDK and Tofino SDEs that would encapsulate finding the libraries and creating targets for them. Then we could simply do a find_package() to obtain the definitions.

PR #35, opened 2022-10-31
Move 'clients' to top-level directory

ffoulkes:
- Moved the `clients` directory from `modules` to the top-level directory. Updated CMake files to reflect change.

- Deleted the `modules` directory, which is no longer needed.


nupuruttarwar (approved):
LGTM

PR #36, opened 2022-10-31
Modify make-all.sh to allow OVS build to be skipped

ffoulkes:
- Implemented `--ovs` and `--no-ovs` options to enable/disable the OVS build. The default is to enable the OVS build. The user may override this by exporting a `BUILD_WITH_OVS` environment variable. A value of "true" enables the OVS build; any other non-empty value disables it.

- Modified the `--clean` option to exit after removing the build and install directories.

### DIFF
--- a/make-all.sh
+++ b/make-all.sh
@@ -2,7 +2,7 @@
 
 # Parse command-line options.
 SHORTOPTS=p:
-LONGOPTS=clean,prefix:,sde-install:,dep-install:,develop,target:
+LONGOPTS=clean,prefix:,sde-install:,dep-install:,develop,target:,ovs,no-ovs
 
 GETOPTS=`getopt -o ${SHORTOPTS} --long ${LONGOPTS} -- "$@"`
 eval set -- "${GETOPTS}"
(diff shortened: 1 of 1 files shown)


make-all.sh
```diff
@@ -67,15 +73,23 @@ fi
...
+if [ -z "${BUILD_WITH_OVS}" -o "${BUILD_WITH_OVS}" = "true" ]; then
+    # Build OVS first.
+    echo "Building OVS"
+    cmake -S ovs -B ${OVS_BUILD} -DCMAKE_INSTALL_PREFIX=${OVS_INSTALL}
+    cmake --build ${OVS_BUILD} -j6 -- V=0
+    OVS_INSTALL_OPTION=-DOVS_INSTALL_DIR=${OVS_INSTALL}
+else
+    echo "Skipping OVS build"
+    OVS_INSTALL_OPTION=
+
```
nupuruttarwar, line 85: remove empty line on 85

nupuruttarwar (approved):
LGTM

PR #42, opened 2022-11-03
Improve setup CMake file

ffoulkes:
- Added an ON_DEMAND option to allow libraries to be built and installed individually.

- Modified the build instructions to run ldconfig after each package is installed if the USE_SUDO option is in effect.

- Add a parameter to the Abseil library build to set the RUNPATH of each library to $ORIGIN.

### DIFF
--- a/setup/CMakeLists.txt
+++ b/setup/CMakeLists.txt
@@ -8,10 +8,12 @@ project(dependencies VERSION 1 LANGUAGES C CXX)
 include(ExternalProject)
 include(CMakePrintHelpers)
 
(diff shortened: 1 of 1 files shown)


nupuruttarwar (approved):
Changes LGTM.. A suggestion to add a readme we discussed on our call specifying the use of ON_DEMAND option and properties like EXCLUDE_FROM_ALL

ffoulkes:
> Changes LGTM.. A suggestion to add a readme we discussed on our call specifying the use of ON_DEMAND option and properties like EXCLUDE_FROM_ALL

An excellent idea.

PR #43, opened 2022-11-03
cmake support part-1

ravi861:
Signed-off-by: Ravi Vantipalli <ravi.vantipalli@intel.com>


infrap4d/CMakeLists.txt
```diff
@@ -2,13 +2,10 @@ cmake_minimum_required(VERSION 3.12)
 
 project(infrap4d VERSION 0.1.0 LANGUAGES C CXX)
 
-set(STRATUM_TDI_BIN_DIR ${STRATUM_SOURCE_DIR}/stratum/hal/bin/tdi)
-set(STRATUM_BFRT_BIN_DIR ${STRATUM_SOURCE_DIR}/stratum/hal/bin/barefoot)
-
 if(DPDK_TARGET)
-    set(INFRAP4D_MAIN ${STRATUM_TDI_BIN_DIR}/dpdk/main.cc)
+    set(INFRAP4D_MAIN $<TARGET_OBJECTS:stratum_dpdk_o>)
```
ffoulkes, line 6: This is is unacceptable. It undoes the mechanism I designed to support infrap4d. We must supply our own main.cc.

infrap4d/CMakeLists.txt
```diff
@@ -19,26 +16,8 @@ target_include_directories(infrap4d PRIVATE ${STRATUM_SOURCE_DIR})
 
 target_link_libraries(infrap4d PRIVATE
     -Wl,--whole-archive
-    stratum
+    stratum_hal_lib_tdi
```
ffoulkes, line 19: This is unacceptable. I need the bulk of stratum as a _static_ library, not a _shared_ library.

clients/gnmi-ctl/CMakeLists.txt
```diff
@@ -17,7 +17,7 @@ add_executable(gnmi-ctl
 
 target_link_libraries(gnmi-ctl
     PUBLIC
-        stratum
+        stratum_hal_lib_tdi
```
ffoulkes, line 20: By linking against the _static_ stratum library, I added only the object files I needed. This adds unneeded files. It may also make us dependent on protobuf files we don't need (assuming that your cli dependencies are correct).

ffoulkes (changes requested):
This breaks the networking recipe.

infrap4d/CMakeLists.txt
ravi861: Modified to use the main separately.

clients/gnmi-ctl/CMakeLists.txt
ravi861: Resolved.

infrap4d/CMakeLists.txt
ravi861: stratum can now be configured to build static or shared libs. Built by default as static.

PR #44, opened 2022-11-03
Fix issues with certian absl targets for latest cmake version

nupuruttarwar:
iLatest cmake version do not expose following internal packages as targets: absl::cord_internal, absl::cordz_info, absl::cordz_handle, absl::cordz_sample_token, absl::cordz_functions and absl::low_level_hash. These packages can be referenced transitively if required.


ffoulkes:
I have verified that this fix works with CMake 3.16.3 on Ubuntu 20.04.

ffoulkes (approved):
LGTM

PR #49, opened 2022-11-10
Update stratum reference point and remove non-inclusive language

nupuruttarwar:


ffoulkes (approved):
LGTM

I like "primary" is cases such as this. I'll have to remember it.

PR #56, opened 2022-11-18
Add krnlmon dependency to stratum for DPDK target

nupuruttarwar:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -98,13 +98,17 @@ endif()
 # Subdirectories #
 ##################
 
+if(DPDK_TARGET)
+    set(WITH_KRNLMON ON)
+endif()
+
+if(WITH_KRNLMON)
+    add_subdirectory(krnlmon)
+endif()
+
 if(WITH_STRATUM)
     add_subdirectory(stratum)
     add_subdirectory(clients)
-endif()
-
-if(DPDK_TARGET)
-add_subdirectory(krnlmon)
 endif()
 
 add_subdirectory(infrap4d)
--- a/stratum/CMakeLists.txt
+++ b/stratum/CMakeLists.txt
@@ -268,12 +268,16 @@ add_library(stratum_target_o OBJECT
     ${STRATUM_SOURCE_DIR}/stratum/hal/lib/tdi/utils.h
 )
 
-target_compile_definitions(stratum_target_o PRIVATE SDE_9_11_0)
-
 target_include_directories(stratum_target_o PRIVATE
     ${STRATUM_INCLUDES}
     ${SDE_INSTALL_DIR}/include
 )
+
+if(WITH_KRNLMON)
+    target_compile_definitions(stratum_target_o PRIVATE KRNLMON_SUPPORT)
+    target_include_directories(stratum_target_o PRIVATE ${KRNLMON_SOURCE_DIR})
+    add_dependencies(stratum_target_o krnlmon)
+endif()
 
 add_dependencies(stratum_target_o stratum_proto)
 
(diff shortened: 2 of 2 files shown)


stratum/CMakeLists.txt
```diff
@@ -424,13 +430,15 @@ elseif(DPDK_TARGET)
     set_property(TARGET target_utils
                  PROPERTY IMPORTED_LOCATION ${LIBTARGET_UTILS})
 
+    add_dependencies(stratum krnlmon)
```
ffoulkes, line 433: 1. You're adding the dependency on krnlmon unconditionally. What happens if `WITH_KRNLMON` is false?
2. Is it necessary to specify the dependency explicitly? From my reading, it looks as if making it a link library does this automatically.

stratum/CMakeLists.txt
```diff
@@ -32,6 +32,10 @@ set(STRATUM_INCLUDES
     ${PROTO_OUT_DIR}
 )
 
+if(DPDK_TARGET)
```
ffoulkes, line 35: 1. The symbol name should be in curly braces (`${KRNLMON_SOURCE_DIR}`).
2. Modern CMake calls for applying properties to individual objects rather than specifying them globally.

stratum/CMakeLists.txt
```diff
@@ -424,13 +430,15 @@ elseif(DPDK_TARGET)
...
 
+    add_dependencies(stratum krnlmon)
     target_link_libraries(stratum PUBLIC
         driver
         bf_switchd_lib
         tdi
         tdi_json_parser
         target_utils
         target_sys
+	krnlmon
```
ffoulkes, line 441: You're adding krnlmon unconditionally.

You need to add it separately:
```
if(WITH_KRNLMON)
    target_link_libraries(stratum PUBLIC krnlmon)
endif()
```

ffoulkes (changes requested):

CMakeLists.txt
```diff
@@ -98,13 +98,13 @@ endif()
 # Subdirectories #
 ##################
 
+if(DPDK_TARGET)
```
ffoulkes, line 101: Instead of basing the dependency on `DPDK_TARGET`, I would recommend implementing a `WITH_KRNLMON` conditional. krnlmon support is not inherently a DPDK property.

Just add the following earlier in the file:
```
if(DPDK_TARGET)
    set(WITH_KRNLMON ON)
endif()
```

stratum/CMakeLists.txt
nupuruttarwar: Added the check and you are right about 2nd. Don't need to specify the dependency explicitly

stratum/CMakeLists.txt
nupuruttarwar: Done

CMakeLists.txt
nupuruttarwar: Good idea, done

ffoulkes (changes requested):

stratum/CMakeLists.txt
```diff
@@ -32,6 +32,10 @@ set(STRATUM_INCLUDES
     ${PROTO_OUT_DIR}
 )
 
+if(WITH_KRNLMON)
+    include_directories(${KRNLMON_SOURCE_DIR})
```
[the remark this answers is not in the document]
nupuruttarwar: Done

stratum/CMakeLists.txt
```diff
@@ -433,6 +442,10 @@ elseif(DPDK_TARGET)
         target_sys
     )
 
+    if(WITH_KRNLMON)
```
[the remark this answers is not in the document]
nupuruttarwar: So, currently we are starting krnlmon_init as part of stratum dpdk_hal, so technically it's integrated for stratum dpdk.
Since stratum thread is blocking thread, we need to figure out how to start these 2 separate threads(stratum, krnlmon).
I will work on this after this drop, but for now it's integrated for stratum dpdk.

ffoulkes (approved):
Approved over my better judgment.

nupuruttarwar:
> Approved against my better judgment.

This will be worked upon as mentioned above to integrate krnlmon into infrap4d instead of stratum

PR #65, opened 2022-11-23
Update readme to export SDE install env variable to sudo user

nupuruttarwar:
Latest SDE use getenv on SDE_INSTALL variable to generate pipeline information dynamically instead of using spec file. Since, infrap4d executable is run in sudo mode, if environment variable is not exported for sudo user, it will failure generating pipeline info and setting the pipeline will fail.


weiqiand (approved):
Looks good to me

PR #67, opened 2022-11-29
Update documentation and dependency build process

nupuruttarwar:
Update grpc cmake command to install absl, protobuf and ssl package as part of grpc build instead of relying on available packages since we strictly require minimum protobuf version 3.18.1. Fix documentation with missing steps


setup/CMakeLists.txt
```diff
@@ -94,9 +94,6 @@ ExternalProject_Add(grpc
     -DCMAKE_INSTALL_PREFIX=${CMAKE_INSTALL_PREFIX}
     -DCMAKE_POSITION_INDEPENDENT_CODE=on
     -DBUILD_SHARED_LIBS=on
-    -DgRPC_ABSL_PROVIDER=package
-    -DgRPC_PROTOBUF_PROVIDER=package
-    -DgRPC_SSL_PROVIDER=package
```
ffoulkes, line 99: 1. If you omit `-DgRPC_SSL_PROVIDER=package`, gRPC will build and install Boring-SSL, which is apparently incompatible with the Tofino SDE.
2. Why not just update the versions of Abseil and Protobuf to be compatible with the version of gRPC? You're sacrificing our ability to patch Abseil or Protobuf independently in case of a security vulnerability.

docs/ipdk-dpdk.md
```diff
@@ -53,7 +53,8 @@ sudo mkdir -p /etc/stratum/
 sudo mkdir -p /var/log/stratum/
```
ffoulkes, line 53: Given the number of (manually-entered) commands, wouldn't it be more convenient for the customer if we put them in a bash script in the `scripts` directory?

setup/CMakeLists.txt
nupuruttarwar: Ohh I was not aware of the first case. On lot of Validation systems, older version of abseil or protobuf were available  and having PROVIDER=package means use the available packages which will cause failure if version requirements are not satisfied. When you say update the version of Abseil and Protobuf, are we expecting the user to do this manually?

docs/ipdk-dpdk.md
nupuruttarwar: Totally agree, will take it as a separate PR

setup/CMakeLists.txt
ravi861: It actually is not a Tofino incompatibility problem. We will run into issues compiling target-utils which houses the tdi_python cpython repo. Python explicitly states that boring-ssl is not supported and incompatible.

ravi861: https://peps.python.org/pep-0644/#boringssl

nupuruttarwar: Added it back

docs/ipdk-dpdk.md
```diff
@@ -9,20 +9,20 @@ This document describes how to build and run IPDK networking recipe on DPDK targ
 ### Build P4 SDE for DPDK
 
 ```bash
-git clone --recursive git@github.com:p4lang/p4-dpdk-target.git p4sde
+   git clone --recursive git@github.com:p4lang/p4-dpdk-target.git p4sde
```
ffoulkes, line 12: I would suggest using https: notation stead of git@ notation.

docs/ipdk-dpdk.md
```diff
@@ -9,20 +9,20 @@ This document describes how to build and run IPDK networking recipe on DPDK targ
...
 ### Build and install infrap4d dependencies
 
 ```bash
-git clone --recursive git@github.com:ipdk-io/networking-recipe.git ipdk.recipe
-cd ipdk.recipe
-export IPDK_RECIPE=`pwd`
-cd $IPDK_RECIPE/setup
-cmake -B build -DCMAKE_INSTALL_PREFIX=<dependency install path> [-DUSE_SUDO=ON]
-cmake --build build [-j<njobs>]
+   git clone --recursive git@github.com:ipdk-io/networking-recipe.git ipdk.recipe
```
ffoulkes, line 20: You generally don't need to indent text in a code block. The change of font plus the gray backdrop make the display text stand out without need for further emphasis.

ffoulkes (approved):

docs/p4rt-ctl.rst
```diff
@@ -303,15 +303,14 @@ services that are available for a p4runtime client.
     ``COUNTER_FLOW``: Refers to the counter ID (generated ID by p4c, refer to bfrt.json file)
     and the counter table index. Format "counter_id=<number>,index=<number>"
 
-Logs and Analysis
-------------------
+Limitations/Note
```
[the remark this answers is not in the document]
nupuruttarwar: Done

nupuruttarwar:
PR is signed

PR #68, opened 2022-12-01
Update refpoint to include new bug fixes

ravi861:
Signed-off-by: Ravi Vantipalli <ravi.vantipalli@intel.com>

### DIFF
--- a/docs/ipdk-tofino.md
+++ b/docs/ipdk-tofino.md
@@ -87,7 +87,7 @@ export SDE_INSTALL=sde/install
 
 ```bash
 cd $IPDK_RECIPE
-./make-all.sh --dep-install /usr/local/ --sde-install $SDE_INSTALL
+./make-all.sh --dep-install /usr/local/ --sde-install $SDE_INSTALL --target=tofino
 ```
 
 ## Run
--- a/docs/ipdk-tofino.md
+++ b/docs/ipdk-tofino.md
@@ -38,7 +38,7 @@ git clone --recursive git@github.com:ipdk-io/networking-recipe.git ipdk.recipe
 
 ### Intel P4Studio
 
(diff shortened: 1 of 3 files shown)


docs/ipdk-tofino.md
```diff
@@ -38,7 +38,7 @@ git clone --recursive git@github.com:ipdk-io/networking-recipe.git ipdk.recipe
 
 ### Intel P4Studio
 
-Get Intel P4Studio SDE version 9.11. The untarred directory is henceforth referred to as **sde**. If using an Intel Tofino Reference platform, please also download the BSP package.
+Get Intel P4Studio SDE. The untarred directory is henceforth referred to as **sde**. If using an Intel Tofino Reference platform, please also download the BSP package.
```
nupuruttarwar, line 41: Say 9.11 or later since TDI was not available before 9.11?

ravi861: At this point 9.11 is not compatible anymore and we have not released 9.12. We have to go back and find a compatible refpoint and tag it for 9.11 compatibility.
Either way, SDE compatibility is to be dealt offline. This document does not need to hold that information.

nupuruttarwar: Then maybe we should change the default target of networking recipe to DPDK?

ravi861: I don't see how it matters but we can do that. In my opinion, there shouldn't be a default target. Let the user explicitly specify what they are interested in building.

nupuruttarwar: Can you make that change - https://github.com/ipdk-io/networking-recipe/blob/main/CMakeLists.txt#L51

nupuruttarwar (approved):
LGTM

PR #76, opened 2022-12-07
Build ovs test controller executable after linking to sidecar library

nupuruttarwar:
Build ovs test controller executable after linking to sidecar library and update ovs submodule reference point


ffoulkes:
Absorbed into PR #77

PR #77, opened 2022-12-07
Fix ovs-testcontroller build and address other CMake issues

ffoulkes:


CMakeLists.txt
```diff
@@ -11,6 +11,9 @@ project(networking-recipe VERSION 0.1 LANGUAGES C CXX)
 include(FindPkgConfig)
 include(CMakePrintHelpers)
 
+# Default: Release with Debug Info
+set(CMAKE_BUILD_TYPE "RelWithDebInfo")
```
nupuruttarwar, line 15: RelWithDbgInfo?

nupuruttarwar (approved):
In general LGTM

CMakeLists.txt
ffoulkes: Release with Debug Information. It sets `-O2 -g -DNDEBUG`.
Release sets `-O3 -DNDEBUG`.
Debug sets `-g`.

nupuruttarwar:
Tested on Fedora 33, able to build successfully and start infrap4d

make-all.sh
```diff
@@ -22,6 +23,9 @@ while true ; do
         # --prefix=<path>
         PREFIX=$2
         shift 2 ;;
+    --debug)
+        BUILD_TYPE="Debug"
```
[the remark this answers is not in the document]
ffoulkes: The default value is RelWithDebInfo, the same as the CMakeLists.txt file.

The other developers want the ability to do debug builds. I took the minimalist approach and implemented just that. We can expand to include Release later (maybe around the time we start adding toolchain support).

PR #79, opened 2022-12-07
Dependency setup script improvements

ffoulkes:
1) Add a check to abort the script if the CMake version is 3.24
   or 3.25. The Protobuf build fails under these versions due to
   a bug in CMake.

2) Modify script to use SHA1 tags for gRPC and Protobuf, to avoid
   unnecessary updates.

3) Specify CMAKE_PREFIX_PATH when building Protobuf and gRPC, to
   ensure that they link against the newly built versions of
   dependent libraries.

4) Download and build v1.2.13 of zlib instead of using the version
   packaged with gRPC, which has known security vulnerabilities.

5) Use the SOURCE_SUBDIR ExternalProject option when building
   Protobuf.

6) Specify the CMAKE_INSTALL_RPATH option when building gRPC,
   glog, and gtest. Use PATCH_COMMAND to update the gRPC
   CMakeLists.txt file to set the INSTALL_RPATH property
   when building the C++ and Python plugins.

### DIFF
--- a/setup/CMakeLists.txt
+++ b/setup/CMakeLists.txt
@@ -8,6 +8,10 @@ project(dependencies VERSION 1 LANGUAGES C CXX)
 include(ExternalProject)
 include(CMakePrintHelpers)
 
+if(${CMAKE_VERSION} VERSION_EQUAL 3.24 OR ${CMAKE_VERSION} VERSION_EQUAL 3.25)
+  message(FATAL_ERROR "Protobuf build is broken in CMake version ${CMAKE_VERSION}")
+endif()
+
 option(ON_DEMAND    "Build targets on demand" OFF)
 option(USE_SUDO     "Use sudo when installing" OFF)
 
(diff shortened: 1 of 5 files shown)


setup/CMakeLists.txt
```diff
@@ -8,6 +8,10 @@ project(dependencies VERSION 1 LANGUAGES C CXX)
 include(ExternalProject)
 include(CMakePrintHelpers)
 
+if(${CMAKE_VERSION} VERSION_EQUAL 3.24 OR ${CMAKE_VERSION} VERSION_EQUAL 3.25)
+  message(FATAL_ERROR "Protobuf build is broken in CMake version ${CMAKE_VERSION}")
```
nupuruttarwar, line 12: Should we specify the recommended version?
Since we have not tested with cmake version greater than 3.25, I am not sure if the bug is fixed in later versions

ffoulkes: Updated message text

nupuruttarwar (approved):
LGTM

PR #83, opened 2022-12-08
Update README file to reflect recent changes

ffoulkes:


nupuruttarwar (commented):
I just realized we should specify the new log location - /var/log/stratum. Can you include this as well in the Readme?  Rest looks good

nupuruttarwar:
Koushik tried building the utilities on a fresh system. 
When building grpc, we provide build options as SSL_PROVIDER=package so we should include installing openssl-devel as part of `### Install basic utilities` in `ipdk-dpdk.md`.
For Fedora :openssl-devel
For Ubuntu: libssl-dev

ffoulkes:
Crikey. There’s more to it than this, isn’t there? Aren’t we now logging to a file instead of stderr by default?

Derek

PR #84, opened 2022-12-09
Integrate krnlmon into infrap4d instead of stratum

nupuruttarwar:
Earlier, krnlmon was tightly coupled into startum and krnlmon_init was called directly from stratum after P4 servers are in listening state. Since this is not a scalable solution, this PR removes direct integration of krnlmon from stratum. Instead, infrap4d spawns 3 threads -
  - krnlmon init
  - krnlmon_shutdown
  - Stratum main. Once P4RT servers are initialized as part of stratum main,it will signal krnlmon_init thread to resume execution.
Similarly, once P4RT servers are stopped, it will signal krnlmon_shutdown thread to resume execution

### DIFF
--- a/infrap4d/CMakeLists.txt
+++ b/infrap4d/CMakeLists.txt
@@ -5,15 +5,15 @@ project(infrap4d VERSION 0.1.0 LANGUAGES C CXX)
 set(STRATUM_TDI_BIN_DIR ${STRATUM_SOURCE_DIR}/stratum/hal/bin/tdi)
 set(STRATUM_BFRT_BIN_DIR ${STRATUM_SOURCE_DIR}/stratum/hal/bin/barefoot)
 
-if(DPDK_TARGET)
+if(WITH_KRNLMON)
+    add_executable(infrap4d infrap4d_main.cc)
+elseif(DPDK_TARGET)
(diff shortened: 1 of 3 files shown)


nupuruttarwar:
Keeping it as draft PR until stratum and krnlmon PR's are merged. Once they are merged, I will modify this PR to update the reference point for these submodules

infrap4d/CMakeLists.txt
```diff
@@ -5,28 +5,37 @@ project(infrap4d VERSION 0.1.0 LANGUAGES C CXX)
 set(STRATUM_TDI_BIN_DIR ${STRATUM_SOURCE_DIR}/stratum/hal/bin/tdi)
 set(STRATUM_BFRT_BIN_DIR ${STRATUM_SOURCE_DIR}/stratum/hal/bin/barefoot)
 
-if(DPDK_TARGET)
+if(WITH_KRNLMON)
+    add_executable(infrap4d infrap4d_main.cc)
```
ffoulkes, line 9: Where is `infrap4d_main.cc`? I don't see it in `infrap4d`, and it's not part of this commit.

ffoulkes (changes requested):

PR #87, opened 2022-12-11
Provide extra control over dependencies setup

ffoulkes:
Provide extra control over dependencies setup

- Implemented a USE_LDCONFIG option to control whether CMake does an
  ldconfig after each install when USE_SUDO is in effect. The option
  defaults to ON for backward compatibility.

  The intention is to allow the user to install to a system directory
  without running ldconfig, if desired.


setup/CMakeLists.txt
```diff
@@ -17,10 +17,13 @@ endif()
 
 option(ON_DEMAND    "Build targets on demand" OFF)
 option(USE_SUDO     "Use sudo when installing" OFF)
+option(USE_LDCONFIG "Use ldconfig when installing" ON)
```
nupuruttarwar, line 20: Should we update documentation to provide explanation for this option

nupuruttarwar: Okay so this is an item for TODO list

nupuruttarwar (approved):
LGTM

PR #88, opened 2022-12-11
Minor setup_env.sh maintenance

ffoulkes:
- Addressed several issues found by shellcheck.
    - Made a couple of edits for readability and consistency.


scripts/dpdk/setup_env.sh
```diff
@@ -29,29 +27,30 @@ export IPDK_RECIPE=$1
 export SDE_INSTALL=$2
 export DEPEND_INSTALL=$3
 
-#Get the OS and Version details
-. $IPDK_RECIPE/scripts/dpdk/os_ver_details.sh
+# Get the OS and Version details
+# shellcheck source=/dev/null
+. "$IPDK_RECIPE/scripts/dpdk/os_ver_details.sh"
```
nupuruttarwar, line 32: Do we need "" around the whole command. We usually put "" around the environment variable for shell check issues, so "$IPDK_RECIPE"/scripts/dpdk/os_ver_details.sh

ffoulkes: This is one of the things shellcheck flagged. I knew about quoting "$OS" (which shellcheck also flagged); this one was a surprise to me.

The issue is [Double quote to prevent globbing and word splitting](https://www.shellcheck.net/wiki/SC2086).

nupuruttarwar (approved):
LGTM

PR #89, opened 2022-12-12
Introduce daemon support for infrap4d for DPDK target

nupuruttarwar:
Add support to run infrap4d as daemon for DPDK target. By default, infrap4d runs in detached mode.
If user wants to run infrap4d in attached mode, use --nodetach option. Add logging information to documentation.

### DIFF
--- a/infrap4d/daemon/daemon-unix.c
+++ b/infrap4d/daemon/daemon-unix.c
@@ -1,0 +1,246 @@
+/*
+ * Copyright (c) 2008, 2009, 2010, 2011, 2012, 2013, 2015 Nicira, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at:
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "daemon.h"
+#include <stdio.h>
+
+#include <sys/types.h>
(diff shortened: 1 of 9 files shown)


infrap4d/daemon/daemon-unix.c
```diff
@@ -0,0 +1,246 @@
...
+#include <unistd.h>
+#include "fatal-signal.h"
+
+#ifdef __linux__
+#define LINUX 1
+#else
+#define LINUX 0
+#endif
+
+/* --detach: Should we run in the background? */
```
ffoulkes, line 36: It just dawned on me: we're linking the daemon code into infrap4d, along with stratum, krnlmon, and a p4 driver. We should ensure that we do not pollute the global namespace in the process.

One solution would be make these files C++, rather than C, and wrap them in namespaces. I'd suggest `inf4ap4d`.

Another solution would be to add a prefix to each global symbol. I would suggest `infrap4d_`.

We do not (necessarily) need to do this for the initial commit, but we should plan to follow up fairly soon.

infrap4d/infrap4d_main.cc
```diff
@@ -21,6 +28,12 @@ int main(int argc, char* argv[]) {
...
 #elif defined(DPDK_TARGET)
+  gflags::ParseCommandLineFlags(&argc, &argv, false);
+  if (FLAGS_detach) {
+      daemonize_start(false);
+      daemonize_complete();
+  }
+
   krnlmon_init();
 
   krnlmon_shutdown();
```
ffoulkes, line 39: The function should end with `return 0;`.

I'm surprised we aren't getting a compiler warning about the function not returning a value. Or are we?

nupuruttarwar: I didn't notice any warning. Added return 0

infrap4d/daemon/daemon-unix.c
nupuruttarwar: You are right, added infrap4d_ prefix to avoid any conflicts for global variables

ffoulkes (approved):
LGTM

PR #90, opened 2022-12-12
Address bandit warnings in p4rt-ctl

ffoulkes:
- Added `# nosec` to assert statements to suppress bandit warnings.


nupuruttarwar (approved):
LGTM

PR #92, opened 2022-12-12
Check if FDB table entry exists before sending INSERT request

nupuruttarwar:
Check if entry was already programmed from previous mac learning using READ P4RT call before issuing a new INSERT call to program the FDB entry. This helps reduce the duplicate entry errors

### DIFF
--- a/ovs-p4rt/ovs_p4rt_session.cc
+++ b/ovs-p4rt/ovs_p4rt_session.cc
@@ -106,7 +106,7 @@ absl::Status GetForwardingPipelineConfig(OvsP4rtSession* session,
   return absl::OkStatus();
 }
 
-absl::StatusOr<ReadResponse> SendReadRequest(OvsP4rtSession* session,
+absl::Status SendReadRequest(OvsP4rtSession* session,
                                              const ReadRequest& read_request) {
   grpc::ClientContext context;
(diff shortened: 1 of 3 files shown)


nupuruttarwar:
This PR is a draft until the GetTable Entry issue given table id and match key is fixed from SDE side. Once this is fixed, similar logic can be applied to ConfigFdbTunnelTableEntry entry. After testing, this PR can be merged

ovs-p4rt/ovs_p4rt_session.cc
```diff
@@ -106,7 +106,7 @@ absl::Status GetForwardingPipelineConfig(OvsP4rtSession* session,
   return absl::OkStatus();
 }
 
-absl::StatusOr<ReadResponse> SendReadRequest(OvsP4rtSession* session,
+absl::Status SendReadRequest(OvsP4rtSession* session,
```
ffoulkes, line 109: The purpose of these changes is to simplify SendReadRequest so it returns just the status code?

You might want to add something to this effect to the commit comment, to provide good documentation for both review and audit trail purposes.

ffoulkes (commented):
LGTM

I see this is a Draft PR. Are we waiting on some event to make it an actual PR?

PR #95, opened 2022-12-14
Fix build error when invoking grpc_cpp_plugin

ffoulkes:
The build failed with the following error message when compiling the C++ protobuf gRPC files:

```
grpc_cpp_plugin: error while loading shared libraries: libprotoc.so.3.18.1.0: cannot open shared object file: No such file or directory
```

The problem turned out to be that, on the system in question, RUNPATH was set to $ORIGIN/../lib and the needed libraries were in $ORIGIN/../lib64.

This changelist modifies the dependencies script to configure the gRPC patch file at build time, using CMAKE_INSTALL_LIBDIR to supply the name of the library subdirectory.


kkongusa (commented):
Keith here, sitting in for Aravind.   Posting my first comment to "test the waters" 8-)

ffoulkes:
> Keith here, sitting in for Aravind. Posting my first comment to "test the waters" 8-)

Here you are, born to be kings, you're the Princes of the Universe. (guitar wails)

5abeel (approved):
lgtm

PR #96, opened 2022-12-14
Build system improvements

ffoulkes:
1) Added support for both lib and lib64 when setting RUNPATH on binary
   executables.

2) Implemented --rpath and --no-rpath options in make-all.sh.

3) Made DEPEND_INSTALL_DIR and SDE_INSTALL_DIR default to the values
   of the DEPEND_INSTALL and SDE_INSTALL environment variables.


5abeel (approved):
lgtm

PR #97, opened 2022-12-14
Distinguish global symbols in infrap4d/daemon

ffoulkes:
- Added a distinguishing prefix to several of function names in the files that were ported from ovs to infrap4d/daemon, to reduce the probability of collisions in the global namespace.

### DIFF
--- a/infrap4d/daemon/daemon-unix.c
+++ b/infrap4d/daemon/daemon-unix.c
@@ -46,7 +46,7 @@ int infrap4d_daemonize_fd = -1;
 static pid_t fork_and_clean_up(void);
 static void daemonize_post_detach(void);
 
-int
+static int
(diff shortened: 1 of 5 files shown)


infrap4d/daemon/daemon-unix.c
```diff
@@ -46,7 +46,7 @@ int infrap4d_daemonize_fd = -1;
 static pid_t fork_and_clean_up(void);
 static void daemonize_post_detach(void);
 
-int
+static int
```
5abeel, line 49: Curious if this is a required change or just a nice-to-have?

5abeel: Understood. I am all in favor of such initiatives!

5abeel (approved):
lgtm

PR #98, opened 2022-12-15
Implement --rpath build option

ffoulkes:
- Modify make-all.sh to support --rpath and --no-rpath options to enable/disable setting the RUNPATH in generated binaries.


5abeel (approved):
LGTM

PR #100, opened 2022-12-17
Fix Fedora33 build issue

ffoulkes:
Append DEPEND_INSTALL_DIR lib64 directory to global link directories. Change required to address the linking issues happening when the networking recipe build process triggered on Fedora33 OS distribution.

Signed-off-by: Venkata Suresh Kumar P <venkata.suresh.kumar.p@intel.com>


ffoulkes (commented):
LGTM

Good catch! I wonder whether we will eventually need to do something similar for `x86_64-gnu-linux`.

vsureshkumarp (approved):
Changes looks good to me.

n-sandeep (approved):
Looks good to me

PR #101, opened 2022-12-18
Overhaul logic to synchronize krnlmon with stratum

ffoulkes:
1) Replaced global pthreads variables with absl::Notification
   objects that are instantiated by infrap4d_main and passed
   as parameters.

2) Implemented target_main() to abstract the call to the main
   function for the selected target.

3) Removed nominal support for TOFINO_TARGET until such time
   as it is needed.

### DIFF
--- a/infrap4d/infrap4d_main.cc
+++ b/infrap4d/infrap4d_main.cc
@@ -2,48 +2,51 @@
 // SPDX-License-Identifier: Apache-2.0
 
 #if defined(DPDK_TARGET)
-#include "stratum/hal/bin/tdi/dpdk/dpdk_main.h"
-#elif defined (TOFINO_TARGET)
-#include "stratum/hal/bin/tdi/tofino/tofino_main.h"
+  #include "stratum/hal/bin/tdi/dpdk/dpdk_main.h"
+#else
+  #error "TDI target type not defined!"
 #endif
 
 extern "C"  {
 #include "daemon/daemon.h"
 }
 
-#include "krnlmon_main.h"
+#include "absl/synchronization/notification.h"
 #include "gflags/gflags.h"
+#include "krnlmon_main.h"
+#include "stratum/glue/status/status.h"
 
 DEFINE_bool(detach, true, "Run infrap4d in attached mode");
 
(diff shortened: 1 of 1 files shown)


infrap4d/infrap4d_main.cc
```diff
@@ -2,48 +2,51 @@
...
       daemonize_complete();
   }
 
-  krnlmon_init();
+  absl::Notification ready_sync;
+  absl::Notification done_sync;
 
-  krnlmon_shutdown();
+  krnlmon_create_main_thread(&ready_sync);
+  krnlmon_create_shutdown_thread(&done_sync);
```
n-sandeep, line 43: Why do we need this shutdown to be called immediately after creating a switchlink thread?

ffoulkes: This function doesn't shut down; it creates a thread that _waits_ for a signal indicating that DpdkHal has shut down. That's why I renamed the function called here from "shutdown" to "create shutdown thread."

Unless I'm greatly mistaken, the existing shutdown thread has a bug in it. I'm pretty sure we should be passing `main_tid` to `switchlink_stop` so the latter can use the pid to stop the main switchlink thread. (Instead, it's using an uninitialized variable as the thread ID.)

I elected not to fix this in the initial Draft PR in order to focus on using dependency inversion to break the tight coupling between DpdkHal and Krnlmon.

nupuruttarwar (commented):
Conversion from pthread mechanism to abseil Notification LGTM, we will need to test this. @n-sandeep, can you please help with testing

n-sandeep:
> Conversion from pthread mechanism to abseil Notification LGTM, we will need to test this. @n-sandeep, can you please help with testing

@nupuruttarwar, Let me discuss with Derek and take it forward.

infrap4d/infrap4d_main.cc
```diff
@@ -2,48 +2,51 @@
 // SPDX-License-Identifier: Apache-2.0
 
 #if defined(DPDK_TARGET)
-#include "stratum/hal/bin/tdi/dpdk/dpdk_main.h"
-#elif defined (TOFINO_TARGET)
```
5abeel, line 6: Why are we getting rid of the TOFINO_TARGET from here?

PR #102, opened 2022-12-19
Update copy_config_files.sh

nupuruttarwar:
Update config file to create /usr/share/target-sys directory where zlog config file needs to be copied


vsureshkumarp (approved):
Changes looks good to me.

n-sandeep (approved):
Looks good to me

PR #116, opened 2023-01-17
Fix protobuf header file dependencies

ffoulkes:
When the networking recipe is built with -j1, compilation of config_monitoring_service.cc fails because openconfig.pb.h has not yet been generated. Addressed this issue by making stratum_hal_lib_common_o dependent on the openconfig_proto target.

When the networking recipe is built in certain parallel build configurations, compilation of dpdk_main.cc fails because gnmi.grpc.pb.h has not yet been generated. Addressed this issue by making stratum_main_o dependent on the gnmi_proto target.


5abeel (approved):
looks good to me

PR #117, opened 2023-01-17
Suppress deprecated-declarations compiler warning

ffoulkes:
Suppress the deprecated-declarations warning when compiling config_monitor_service.cc. The warning appears to be interfering with Coverity analysis.


ffoulkes:
> I understand that we want to suppress this warning because it is in Stratum code and not the code we wrote?

I want to make the warning go away because it appears to be causing the Coverity scan to fail outright.

I looked for an alternative method to using the deprecated method and was referred to a Go module. (Not much help.) I then checked the latest Stratum code. (Not fixed there) The lowest-impact change I could think of was to suppress the deprecated-method warning for just this file.

5abeel:
Approved

PR #122, opened 2023-01-20
Fix ovs-p4ctl tls mode to follow similar logic as CredentialsManager

5abeel:
Signed-off-by: Sabeel Ansari <sabeel.ansari@intel.com>


vsureshkumarp (approved):
Looks good me

PR #128, opened 2023-01-24
Address issues with CMAKE_BUILD_TYPE

ffoulkes:
The CMAKE_BUILD_TYPE definition in the top-level CMakeLists.txt file wasn't working as expected. It turns out that it has to be defined _before_ the project() command or it has no effect.

- Moved the definition of CMAKE_BUILD_TYPE before the project() command.
- Defined a new KLOCWORK command-line option, which defaults to OFF.
- Modified CMAKE_BUILD_TYPE to default to "Debug" if the KLOCWORK option is enabled or "RelWithDebInfo" if is is disabled.

When building in Debug mode, CMake suppresses the definition of the NDEBUG flag, thereby enabling any assert() statements in the code. We do this so Klocwork takes variable references in the assertions into account when doing static analysis.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,14 +6,20 @@
 
 cmake_minimum_required(VERSION 3.5)
 
+option(KLOCWORK "Configure for Klocwork scan" OFF)
+
+# CMAKE_BUILD_TYPE must be set before the first project() command.
+if(KLOCWORK)
+  set(CMAKE_BUILD_TYPE "Debug" CACHE STRING "Type of build to perform")
+else()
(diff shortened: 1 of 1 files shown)


ravi861 (commented):
Would this override the value if the value is set from command line?
cmake .. -DCMAKE_BUILD_TYPE=Release

ffoulkes:
> Would this override the value if the value is set from command line? cmake .. -DCMAKE_BUILD_TYPE=Release

The command line takes precedence over the definitions in the CMakeLists file.

Namrata-intel (approved):

vsureshkumarp (approved):
Changes looks good to me.

PR #135, opened 2023-01-26
Overhaul logic to synchronize krnlmon with stratum

ffoulkes:
1) Replaced global pthreads variables with `absl::Notification` objects that are instantiated by `infrap4d_main` and passed as parameters.

2) Implemented `target_main()` to abstract the call to the main function for the selected target, making the code easier to read and maintain.

3) Removed tentative support for TOFINO_TARGET until it is actually needed.

4) Updated krnlmon and stratum reference points.

This changelist must be coordinated with changes in the stratum and krnlmon submodules.


ffoulkes:
Rebased on current HEAD and updated the submodule pointers (which will need to be updated again after the krnlmon and stratum submodules are updated).

5abeel (approved):
LGTM

PR #136, opened 2023-01-27
CMake script refinements

ffoulkes:
- Renamed the `GOOGLE_SOURCE_DIR` variable to `GOOGLEAPIS_SOURCE_DIR`, to emphasize its association with the `googleapis` directory.

- Renamed the `PROTO_OUT_DIR` variable to `PB_OUT_DIR`, to emphasize its association with the `pb.out` directory.

- Renamed the `stratum` target to `stratum_static`, to emphasize the fact that the library is static by design. This reserves the `stratum` target name for possible use for a shared library.

- Made `stratum_lib_o` dependent on the `gnmi_proto` target, to ensure that `gnmi.pb.h` is generated before we attempt to compile `credentials_manager.cc`.

- Changed networking-recipe project version to 23.01, to match external numbering.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,7 +15,7 @@ else()
   set(CMAKE_BUILD_TYPE "RelWithDebInfo" CACHE STRING "Type of build to perform")
 endif()
 
-project(networking-recipe VERSION 0.1 LANGUAGES C CXX)
+project(networking-recipe VERSION 23.01 LANGUAGES C CXX)
 
 include(CMakePrintHelpers)
 include(FindPkgConfig)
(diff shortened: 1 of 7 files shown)


CMakeLists.txt
```diff
@@ -39,9 +39,9 @@ set(OVS_SOURCE_DIR "${CMAKE_SOURCE_DIR}/ovs/ovs" CACHE PATH
 set(SDE_INSTALL_DIR "$ENV{SDE_INSTALL}" CACHE PATH
     "SDE install directory")
 
-set(PROTO_OUT_DIR "${CMAKE_BINARY_DIR}/pb.out" CACHE PATH
+set(PB_OUT_DIR "${CMAKE_BINARY_DIR}/pb.out" CACHE PATH
```
5abeel, line 42: Would `PROTOBUF_OUT_DIR` be a better name?

5abeel (approved):
Just a name suggestion consideration. Rest looks good.

CMakeLists.txt
ffoulkes: `PB_OUT_DIR` tracks the directory name, which is `pb.out`.

There are two kinds of Protobuf files: files written in the Protobuf interface language, which have the file type `.proto`, and language-specific files generated by the Protobuf compiler. For C++, these files have file types `.pb.cc` and `.pb.h`. The files written to the `pb.out` directory are the latter.

`proto` refers to files that are inputs to the Protobuf compiler. `pb` refers to files that are outputs of the Protobuf compiler.

5abeel: Ack.
Rest looks good.

PR #138, opened 2023-01-28
Partition gnmi_proto and openconfig_proto library builds

ffoulkes:
Moved the logic to build the `gnmi_proto` and `openconfig_proto` libraries from `stratum/proto/CMakeLists.txt` to CMakeLists.txt files in the `stratum/proto/gnmi` and `stratum/proto/openconfig` subdirectories.

This is a cleanup change to improve the modularity and maintainability of the build system.

### DIFF
--- a/stratum/proto/CMakeLists.txt
+++ b/stratum/proto/CMakeLists.txt
@@ -4,8 +4,6 @@
 # SPDX-License-Identifier: Apache 2.0
 #
 
-cmake_minimum_required(VERSION 3.5)
-
 find_package(Protobuf REQUIRED)
 
 ###############################
--- a/stratum/proto/CMakeLists.txt
+++ b/stratum/proto/CMakeLists.txt
@@ -1,11 +1,12 @@
 # Builds protobuf object libraries
+#
+# Copyright 2022-2023 Intel Corporation
+# SPDX-License-Identifier: Apache 2.0
+#
 
 cmake_minimum_required(VERSION 3.5)
 
 find_package(Protobuf REQUIRED)
-if(NOT Protobuf_FOUND)
-    message(SEND_ERROR "Error: Protobuf package not found")
-endif()
 
 ###############################
 # Get path to GRPC C++ plugin #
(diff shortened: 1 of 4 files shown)


stratum/proto/CMakeLists.txt
```diff
@@ -1,11 +1,12 @@
 # Builds protobuf object libraries
+#
+# Copyright 2022-2023 Intel Corporation
+# SPDX-License-Identifier: Apache 2.0
+#
 
 cmake_minimum_required(VERSION 3.5)
 
 find_package(Protobuf REQUIRED)
-if(NOT Protobuf_FOUND)
```
5abeel, line 6: I gather this is being removed because we are building and adding dependency, but would it make sense to leave the error this message in case of build errors that might be hard to catch otherwise?

stratum/proto/gnmi/CMakeLists.txt
```diff
@@ -0,0 +1,50 @@
+# Builds gNMI Protobuf library
+#
+# Copyright 2022-2023 Intel Corporation
+# SPDX-License-Identifier: Apache 2.0
+#
+
+cmake_minimum_required(VERSION 3.12)
```
5abeel, line 7: The other CMake file has a 3.5 minimum version requirement. Better to keep it consistent with whatever our minimum requirement is.

stratum/proto/openconfig/CMakeLists.txt
```diff
@@ -0,0 +1,90 @@
+# Builds OpenConfig Protobuf library
+#
+# Copyright 2022-2023 Intel Corporation
+# SPDX-License-Identifier: Apache 2.0
+#
+
+cmake_minimum_required(VERSION 3.12)
```
5abeel, line 7: Same comment as above re cmake version number

5abeel (changes requested):

stratum/proto/gnmi/CMakeLists.txt
ffoulkes: I disagree. The parent CMakeLists.txt file specifies a minimum version of 3.5, which this file inherits. I'm not aware of any features in the file that require a higher version number, which is why I omitted the command. (Bear in mind, the commands in this file were moved here from the parent file.)

We should be _removing_ unnecessary cmake_minimum_required() commands, not adding them!

stratum/proto/openconfig/CMakeLists.txt
ffoulkes: I disagree. The parent CMakeLists.txt file specifies a minimum version of 3.5, which this file inherits. I'm not aware of any features in the file that require a higher version number, which is why I omitted the command. (Bear in mind, the commands in this file were moved here from the parent file.)

We should be _removing_ unnecessary cmake_minimum_required() commands, not adding them!

stratum/proto/CMakeLists.txt
ffoulkes: I removed the explict check because it's dead code. The REQUIRED parameter on the find_package() command guarantees that it will fail (with a suitable error message) if the Protobuf package is not found.

stratum/proto/gnmi/CMakeLists.txt
5abeel: So, this can be removed and it will run with the inherited cmake version 3.5 requirement?

ffoulkes: Correct.

5abeel: Can we add a `TODO` comment to mark it as potential cleanup in the future? Approving it now, but hopefully this doesnt get lost and forgotten about

5abeel (approved):

stratum/proto/gnmi/CMakeLists.txt
ffoulkes: I've promoted `stratum/CMakeLists.txt` to version 3.12 (the suggested base version for "Modern CMake") and removed the `cmake_minimum_version()` commands from all the subordinate CMakeLists.txt files. This is not a definitive solution, but it gives us a degree of consistency in this subtree, at least for the time being.

PR #142, opened 2023-02-28
Update networking-recipe for ES2K target

ffoulkes:
Added support for the Intel Infrastructure Processing Unit (IPU) E2000 ASIC.


.github/workflows/build_and_test.yml
```diff
@@ -19,7 +19,7 @@ jobs:
 
   build_and_test:
     timeout-minutes: 100
-    runs-on: ubuntu-20.04
+    runs-on: gasp
```
5abeel, line 22: Probably a question for @mgaravin -- how would this work for MEV/ES2K target? Would CI builds still be supported through the Github runners?

nupuruttarwar (approved):
LGTM

.github/workflows/build_and_test.yml
ffoulkes: No clue. We'll have to wait for the CI/CD boffins to weigh in.

mgaravin: Yes, we have Intel shared GitHub runners provided by 1source team! I'm trying to make that work, if not i have to add dedicated "self-hosted" runner and execute the github actions.

Pls. note: Self-host runner can't run all opensource github actions ( as few actions as expected to run only in opensource environment)! have to check and do modifications in innersource based on our need!

mgaravin: And, probably, we should not upstream innersource github actions to opensource, as it was customized to run only on innersource environment

ffoulkes: I have backed out the changes to the github actions

ffoulkes:
We are currently blocked, waiting for approval to open-source the MEV changes.

ffoulkes (commented):
Reviewed in order to create a summary of changes made for ES2K.

Added a number of comments along the way.

PR #144, opened 2023-05-19
Backport ES2K changes to ipdk-io

ffoulkes:


nupuruttarwar (commented):
Mirroring feature is supported only for E2100 right now but there are changes in common files.
for example: mvlut changes in p4rt-ctl.in

n-sandeep:
> Mirroring feature is supported only for E2100 right now but there are changes in common files. for example: mvlut changes in p4rt-ctl.in

In case p4runtime clinet, server should reply back with an valid error message similar to "feature not supported". For now i believe server throws an error message because LUT_manager is not available on server,. This can be enhanced in later fixes.

5abeel (approved):
LGTM

n-sandeep:
This PR is also trying to update submodles as well with the SHA of open PR in each submodule, AFAIK once the submodule PR's are merged it will generate a new SHA.
So, a new patch need to be raised to have correct submodule SHA, right?

ffoulkes:
> Mirroring feature is supported only for E2100 right now but there are changes in common files. for example: mvlut changes in p4rt-ctl.in

We don't want to have to support target-specific versions of the client utilities. The most straightforward solution is to have the target return an error if it doesn't support the request.

ffoulkes:
> This PR is also trying to update submodles as well with the SHA of open PR in each submodule, AFAIK once the submodule PR's are merged it will generate a new SHA. So, a new patch need to be raised to have correct submodule SHA, right?

My usual practice is to merge the submodules, pull them (so my commit IDs match those on the server), update the submodule references in the superproject, and merge that. That should ensure that we don't push stale commit IDs to the main branch.

nupuruttarwar (approved):
LGTM

PR #150, opened 2023-06-02
Create scripts to patch DPDK and ES2K SDEs

ffoulkes:
- Implement scripts to create a TARGET file in the SDE `share` directory that specifies the target type. Bazel uses this file to verify the SDE type when building ES2K targets.

- Implement scripts to patch several SDE target libraries so `infrap4d` can be loaded without requiring the `LD_LIBRARY_PATH` environment variable. Note that P4 Control Plane needs to be built with the RPATH option enabled for this to work.

### DIFF
--- a/scripts/dpdk/set-dpdk-sde-target.sh
+++ b/scripts/dpdk/set-dpdk-sde-target.sh
@@ -1,0 +1,31 @@
+# Copyright 2022-2023 Intel Corporation
+# SPDX-License-Identifier: Apache 2.0
+#
+# Creates a share/TARGET file that identifies this as an ES2K SDE.
+# This allows Bazel (or any other tool) to determine the SDE type.
+#
+
+# Abort on error.
+set -e
+
+# Error check.
+if [ -z "${SDE_INSTALL}" ]; then
+    echo "SDE_INSTALL not defined!"
+    exit 1
+fi
+
+print_target_type() {
+    read TYPE
+    echo "Target type is '$TYPE'"
+}
+
+# Create TARGET file.
+filename="${SDE_INSTALL}/share/TARGET"
+if [ ! -e "${filename}" ]; then
+    echo "Creating ${filename}"
+    echo "DPDK" > ${filename}
+    echo "TARGET file created"
+else
+    echo "TARGET file already exists"
+    print_target_type < ${filename}
+fi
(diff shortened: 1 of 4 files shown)


scripts/dpdk/set-dpdk-sde-target.sh
```diff
@@ -0,0 +1,31 @@
+# Copyright 2022-2023 Intel Corporation
+# SPDX-License-Identifier: Apache 2.0
+#
+# Creates a share/TARGET file that identifies this as an ES2K SDE.
```
nupuruttarwar, line 4: DPDK SDE?

nupuruttarwar (commented):
LGTM logically

We should add documentation in dpdk and es2k guide about how to compile the recipe with rpath enabled and introduced an option --with-rpath  in setup_env.sh for both dpdk and es2k scripts. If enabled, these patches will be applied and LD_LIBRARY_PATH will not be set. If this isn't heavily tested we can say this is an experimental support.

Thoughts?

scripts/dpdk/set-dpdk-sde-target.sh
ffoulkes: Aw, nuts..

ffoulkes:
> LGTM logically
> 
> We should add documentation in dpdk and es2k guide about how to compile the recipe with rpath enabled and introduced an option --with-rpath in setup_env.sh for both dpdk and es2k scripts. If enabled, these patches will be applied and LD_LIBRARY_PATH will not be set. If this isn't heavily tested we can say this is an experimental support.
> 
> Thoughts?

There's already an `--rpath` option in `make-all.sh`. It's been in the innersource code base for a couple of months (but not in the make-all help text.)

I recently pushed a PR that changes the CMake default to building with RPATH enabled. The make-all script defers to CMake unless the user explicitly specifies --rpath or --no-rpath, in which case it overrides the CMake default.

My thinking was that the patch files are strictly voluntary ("don't monkey with the SDE without my express request"). You seem to be in favor of applying them automatically ("set up the system as optimally as possible"). Both are valid points of view. Perhaps we should discuss this with a larger group.

ffoulkes:
I recently learned of a `chrpath` command that changes only the RUNPATH. This might be preferable to `patchelf`.

PR #151, opened 2023-06-02
Enable RPATH support by default

ffoulkes:
cmake build system:

- Change default value of SET_RPATH option from OFF to ON.

make-all.sh:

- Change default value of RPATH option from OFF to undefined. --rpath and --no-rpath cause the appropriate SET_RPATH argument to be specified on the cmake command line, overriding the default in the cmake command file.

- Add --rpath and --no-rpath options to the help text.

config-cross-recipe.sh:

- Add support for --rpath and --no-rpath options, with the same semantics as make-all.sh.

- Refactor code to display output for the dry-run option into a print_cmake_params() function.

- Separate help text into Paths and Options sections.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -86,7 +86,7 @@ unset(_target_action)
 # Feature toggles #
 ###################
 
-option(SET_RPATH    "Set RPATH in libraries and executables" OFF)
+option(SET_RPATH    "Set RPATH in libraries and executables" ON)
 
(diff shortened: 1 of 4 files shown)


nupuruttarwar (approved):
LGTM

ffoulkes:
Note that the ES2K cross-compilation script currently enables RPATH, so there will be no change in default behavior (but the user will now be able to disable it if they wish).

PR #158, opened 2023-06-08
enable ovs and stratum unit tests for dpdk target

mgaravin:
1. enabled OVS unit test and archive test results 
2. enabled stratum unit tests for DPDK target and archive test results
3. todo: krnlmon ut


.github/workflows/build_and_test.yml
```diff
@@ -71,17 +73,88 @@ jobs:
...
+              sudo rm -rf ~/.local/bin/cmake || true
+              # removing cmake 3.25 and reinstalling 3.20, as build is failing with cmake 3.25
+              wget https://cmake.org/files/v3.20/cmake-3.20.2-linux-x86_64.sh -O ${GITHUB_WORKSPACE}/cmake-3.20.2-linux-x86_64.sh
+              chmod a+x ${GITHUB_WORKSPACE}/cmake-3.20.2-linux-x86_64.sh
+              sudo mkdir -p /opt/rh/cmake
+              cd ${GITHUB_WORKSPACE} && sudo sh cmake-3.20.2-linux-x86_64.sh --prefix=/opt/rh/cmake --skip-license
+              sudo ln -s /opt/rh/cmake/bin/cmake /usr/local/bin/cmake
+              cmake --version
 
+      - name: 'Checkout Networking Recipe'
```
ffoulkes, line 137:
```suggestion
      - name: 'Check Out Networking Recipe'
```
- _checkout_ (one word) is a noun or adjective. "I went to the checkout counter."
- _check out_ (two words) is prepositional verb (verb+preposition used as a verb). "I need to check out before I leave."

.github/workflows/build_and_test.yml
```diff
@@ -71,17 +73,88 @@ jobs:
...
+        uses: actions/checkout@v3
+        with:
+          submodules: 'recursive'
+          path: ipdk.recipe
+          
+      - name: 'Run OVS unit tests'
```
ffoulkes, line 143: The OVS unit tests do not need to be run on every commit. OVS is a third-party package that rarely changes. If we're going to spend the time, there are other checks that are far more valuable than this.

ffoulkes (commented):
The OVS unit tests should be run on a cadence, on demand, or (if feasible) on a PR that modifies the OVS repository. It's third-party package the we modify infrequently.

.github/workflows/build_and_test.yml
mgaravin: Removed OVS unit test from this flow - will add individual action based on schedule/on demand builds

.github/workflows/build_and_test.yml
mgaravin: Fixed!

.github/workflows/build_and_test.yml
```diff
@@ -71,17 +73,37 @@ jobs:
...
       - name: 'Execute dpdk target build'
         run: |
              cd ${GITHUB_WORKSPACE}/ipdk.recipe
              export IPDK_RECIPE=${GITHUB_WORKSPACE}/ipdk.recipe
              export SDE_INSTALL=${GITHUB_WORKSPACE}/install
              export DEPEND_INSTALL=${GITHUB_WORKSPACE}/deps
              source ./scripts/dpdk/setup_env.sh $IPDK_RECIPE $SDE_INSTALL $DEPEND_INSTALL
-             ./make-all.sh --target=dpdk
-
+             ./make-all.sh --target=dpdk  --prefix=${GITHUB_WORKSPACE}/deps
```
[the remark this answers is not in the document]
ffoulkes: For example:
```
INSTALL
├── deps
│   ├── bin
│   ├── include
│   ├── lib
│   └── share
├── dpdk-target
│   ├── bin
│   ├── include
│   ├── lib
│   ├── sbin
│   └── share
├── es2k-target
│   ├── bin
│   ├── include
│   ├── lib
│   ├── sbin
│   └── share
└── ovs
    ├── bin
    ├── etc
    ├── include
    ├── lib
    ├── sbin
    ├── share
    └── var
```

ffoulkes (approved):
LGTM

This is okay as it is. I've suggested a possible future improvement.

.github/workflows/build_and_test.yml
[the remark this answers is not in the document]
mgaravin: How are we going to get es2k-sde, as it would be developed in innersource model?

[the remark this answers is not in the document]
mgaravin: Thanks Derek for the suggestions!
can i consider the suggestions in next PR?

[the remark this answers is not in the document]
ffoulkes: > How are we going to get es2k-sde, as it would be developed in innersource model?

The SDE will need to be prebuilt from innersource, packaged, and somehow made available to CI.

In what context are GitHub actions run? Is there a special user account associated with the CI workflows? If so, perhaps we could create a private repository on ipdk-io, archive versions of the binary package there, and grant the CI account read access to the repository. The only gotcha I can think is providing a mechanism to specify the version of the binary package the workflow should fetch.

Come to think of it, we might need to test periodically against different versions of the SDE to verify backward compatibility.

Yikes. The problem space gets larger and more complicated every time I think about it.

ffoulkes, line 105:
```
./make-all.sh --target=dpdk  --prefix=${GITHUB_WORKSPACE}/deps 
```
This installs the dpdk target code into the dependencies directory. I'm recommending _against_ doing this, for the reasons explained in the comment above.

ffoulkes (approved):
I don't agree with your approach, but it should work. Approved.

.github/workflows/build_and_test.yml
mgaravin: ahh, i have overlooked the comment, fixed in latest commit! Thanks!

.github/workflows/build_and_test.yml
```diff
@@ -71,17 +73,37 @@ jobs:
              export IPDK_RECIPE=`pwd`
              cd $IPDK_RECIPE/setup
              cmake -B build -DCMAKE_INSTALL_PREFIX=${GITHUB_WORKSPACE}/deps
-             cmake --build build -j6
+             cmake --build build -j6
```
nupuruttarwar, line 76: remove extra space?

.github/workflows/build_and_test.yml
```diff
@@ -71,17 +73,37 @@ jobs:
              export IPDK_RECIPE=`pwd`
              cd $IPDK_RECIPE/setup
              cmake -B build -DCMAKE_INSTALL_PREFIX=${GITHUB_WORKSPACE}/deps
-             cmake --build build -j6
+             cmake --build build -j6   
 
+      - name: 'Run Stratum unit tests'
+        run: |
+             cd /tmp
+             curl -o bazel-4.2.3-installer-linux-x86_64.sh https://github.com/bazelbuild/bazel/releases/download/4.2.3/bazel-4.2.3-installer-linux-x86_64.sh
```
nupuruttarwar, line 81: Can we add a log here saying "Installing Bazel"?

.github/workflows/build_and_test.yml
```diff
@@ -71,17 +73,37 @@ jobs:
              export IPDK_RECIPE=`pwd`
              cd $IPDK_RECIPE/setup
              cmake -B build -DCMAKE_INSTALL_PREFIX=${GITHUB_WORKSPACE}/deps
-             cmake --build build -j6
+             cmake --build build -j6   
 
+      - name: 'Run Stratum unit tests'
```
5abeel, line 78: Why run Stratum unit tests in the context of networking-recipe repository?

If a new PR is raised in Stratum submodule, we should run the unit tests on the open PR and/or on merge. At this point networking-recipe is still pointing to Stratum's (HEAD-1) SHA.
Networking-recipe is a follow-on separate commit and this test will only get trigger when we try to update the submodule

I think a better design/arch choice would be to do this in Stratum repo

.github/workflows/build_and_test.yml
```diff
@@ -71,17 +73,37 @@ jobs:
...
+             echo "DPDK" > ${GITHUB_WORKSPACE}/install/share/TARGET
+             xargs -a ${GITHUB_WORKSPACE}/ipdk.recipe/.github/dpdk-tests.txt bazel test --define target=dpdk
+             cd bazel-testlogs/stratum
+             zip stratum_ut_logs.zip `find . -name test.xml -print`
+                          
+      - name: 'Upload stratum-ut-result artifact'
+        uses: actions/upload-artifact@v3
+        with:
+          name: stratum-ut-result
+          path: ipdk.recipe/stratum/stratum/bazel-testlogs/stratum/stratum_ut_logs.zip
```
5abeel, line 96: How do we access these logs? Do they need to get versioned to differentiate between subsequent runs?

.github/workflows/build_and_test.yml
ffoulkes: Interesting point.

.github/workflows/build_and_test.yml
ffoulkes: I believe the intention is to feed the test results to a dashboard. I don't know if that includes providing access to the individual test logs.

.github/workflows/build_and_test.yml
ffoulkes: As a matter of practice, we should trim trailing white space before committing the file.

.github/workflows/build_and_test.yml
mgaravin: @ffoulkes : I think, we had a similar discussion and you mentioned Stratum tests target specific. If i remember correctly,  i too suggested to run stratum tests under stratum repo.

.github/workflows/build_and_test.yml
mgaravin: Yes, for now, we need some modifications to display in dashboard, but test results are avilable under artifacts section of each GitHub run

image

.github/workflows/build_and_test.yml
```diff
@@ -71,17 +73,37 @@ jobs:
...
              cmake -B build -DCMAKE_INSTALL_PREFIX=${GITHUB_WORKSPACE}/deps
-             cmake --build build -j6
+             cmake --build build -j6   
 
+      - name: 'Run Stratum unit tests'
+        run: |
+             cd /tmp
+             curl -o bazel-4.2.3-installer-linux-x86_64.sh https://github.com/bazelbuild/bazel/releases/download/4.2.3/bazel-4.2.3-installer-linux-x86_64.sh
+             chmod +x bazel-4.2.3-installer-linux-x86_64.sh
+             ./bazel-4.2.3-installer-linux-x86_64.sh --user
```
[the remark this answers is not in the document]
mgaravin: By default, GitHub action would exit if installation fails, but it is highly depends on "bazel-4.2.3-installer-linux-x86_64.sh" behavior!

.github/workflows/build_and_test.yml
mgaravin: Yep, i updated my vscode settings!

.github/workflows/build_and_test.yml
mgaravin: done!

5abeel (approved):
LGTM

PR #162, opened 2023-06-11
Revise target selection logic

ffoulkes:
- Implement TDI_TARGET cmake option to specify the build target.
- Move selection logic to an include file, for modularity.
- Provide backward compatibility with the DPDK_TARGET, ES2K_TARGET, and TOFINO_TARGET options.
- Set TARGETTYPE to the type of target selected (DPDK, ES2K, etc.).
- Update make-all.sh to use TDI_TARGET instead of legacy option.


ffoulkes:
The current implementation defines a separate command-line option for each possible target (`DPDK_TARGET`, `ES2K_TARGET`, `TOFINO_TARGET`). This approach does not scale well.

The new implementation defines a single command-line option that applies to all targets (`TDI_TARGET=<target>`).

For backward compatibility, the new implementation continues to support the existing target-specific options. It writes a message to the console stating that the option is deprecated, but continues to accept it:
```
$ cmake -B build -DTOFINO_TARGET=ON
TOFINO_TARGET option is deprecated; use TDI_TARGET=TOFINO instead.
Building TOFINO_TARGET
```

It also checks for invalid combinations of options:
```
$ cmake -B build -DTOFINO_TARGET=ON -DTDI_TARGET=ES2K
TOFINO_TARGET option is deprecated; use TDI_TARGET=TOFINO instead.
CMake Error at cmake/select_tdi_target.cmake:83 (message):
  TOFINO_TARGET and TDI_TARGET=ES2K both specified!
-- Configuring incomplete, errors occurred!
```

and:
```
$ cmake -B build -DDPDK_TARGET=on -DTOFINO_TARGET=on
DPDK_TARGET option is deprecated; use TDI_TARGET=DPDK instead.
TOFINO_TARGET option is deprecated; use TDI_TARGET=TOFINO instead.
CMake Error at cmake/select_tdi_target.cmake:55 (message):
  More than one legacy target option specified!
-- Configuring incomplete, errors occurred!
```
The plan is to increase the severity of the message at intervals of, say, one month:
- From `NOTIFICATION` to `WARNING` or `DEPRECATION`
- Then to `FATAL_ERROR`
- Then drop all special processing of the legacy options.

The latter will result in a message such as the following:
```
-- Generating done
CMake Warning:
  Manually-specified variables were not used by the project:

    DPDK_TARGET

-- Build files have been written to: /home/dfoster/latest/build
```

cmake/select_tdi_target.cmake
```diff
@@ -0,0 +1,109 @@
...
+#
+# Issues deprecation messages for legacy target options that are enabled.
+#
+# Sets TARGETFLAG and the corresponding Boolean target option.
+# Unsets all other Boolean target options.
+#
+
+set(TDI_TARGET "" CACHE STRING "TDI Target Type")
+
+set(_valid_tdi_targets DPDK ES2K TOFINO)
```
5abeel, line 17: Is this where we would add more targets if supported in future?

5abeel: If it is a simple addition here, thats good. If more places need manual editing, will need to document (or add a comment here) that other changes may be needed elsewhere

5abeel (commented):
I like this change, since it can scale easily for more targets in the future

cmake/select_tdi_target.cmake
ffoulkes: > Is this where we would add more targets if supported in future?

Yes, that's the plan.  If it turns out we need the information elsewhere (which I don't expect to happen), we can rename the symbol to `VALID_TDI_TARGETS` and remove the command that unsets it at the end of the module.

nupuruttarwar (approved):
LGTM

satish153 (commented):
Currently, the user has to input ON/OFF for a target. With this change, the user has to input the exact target name. There is a chance of typing in an incorrect target name. Eg: 'dpdk' instead of DPDK.
Will listing the currently supported targets be helpful when the user inputs incorrect command line options?

ffoulkes:
> Will listing the currently supported targets be helpful when the user inputs incorrect command line options?

Good idea.

satish153 (approved):
LGTM

PR #156, opened 2023-06-07
Stratum cmake support

ffoulkes:
- Move basic cmake build support from the superproject into the stratum project. Protobuf generation and compilation is still part of the networking-recipe.
- Replace our implementation of FindUnwind with one taken from the glog project. This resolves a conflict between the two versions. Add copyright notice, SPDX license identifier, and link to the license file in the glog project.
- Change target library references to unwind::unwind.

PR #163, opened 2023-06-11
Standardize cmake version

ffoulkes:
- Specify `cmake_minimum_required(VERSION 3.15)` in all CMakeLists.txt files that are the root of a project().

- Remove `cmake_minimum_required()` commands from all CMakeLists.txt files that are not the root of a project().

### DIFF
--- a/krnlmon/CMakeLists.txt
+++ b/krnlmon/CMakeLists.txt
@@ -4,8 +4,6 @@
 # SPDX-License-Identifier: Apache 2.0
 #
 
-cmake_minimum_required(VERSION 3.15)
-
 # Path to SAI source directory (for krnlmon)
 set(SAI_SOURCE_DIR "${CMAKE_CURRENT_SOURCE_DIR}/SAI" CACHE PATH
     "Path to SAI source directory")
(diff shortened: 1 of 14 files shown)


krnlmon/CMakeLists.txt
```diff
@@ -4,8 +4,6 @@
 # SPDX-License-Identifier: Apache 2.0
 #
 
-cmake_minimum_required(VERSION 3.15)
```
5abeel, line 7: Don't we need a `cmake_minimum_required` in krnlmon as we do with the ovs/CMakeLists.txt and stratum/CMakelists, since they are submodules?

ffoulkes: The `cmake_minimum_required()` for krnlmon is in `krnlmon/krnlmon/CMakeLists.txt`, which is where the `project()` command is. It's the top-level directory in the `krnlmon` repository.

I couldn't do this for `stratum` because cmake support stops at repository boundary. This will change when we move cmake support into the Stratum repository (Stratum PR https://github.com/ipdk-io/stratum-dev/pull/77 and networking-recipe PR https://github.com/ipdk-io/networking-recipe/pull/156).

5abeel (approved):
LGTM

nupuruttarwar (approved):
LGTM

PR #166, opened 2023-06-12
Fix documentation with correct information

nupuruttarwar:


docs/guides/es2k-guide.md
```diff
@@ -129,10 +129,8 @@ After installing ATE Kernel on HOST machine, install the following drivers
 to bind the network devices (netdevs) to the E2100 target.
 
 ```bash
+modprobe auxiliary 
 modprobe idpf
-modprobe auxiliary
-modprobe iecm
```
5abeel, line 134: Confirming that we do not need iecm and iavf?

docs/guides/security-guide.md
```diff
@@ -64,7 +51,7 @@ COMMON_NAME=localhost ./generate-certs.sh
 ```
 
 * Copy the generated ca.crt, stratum.crt, and stratum.key to the server
-  running InfraP4D
+  running infraP4D
```
5abeel, line 54: Need to be all in lowercase `infrap4d`

5abeel (commented):
Some minor edits needed, rest looks good

docs/guides/es2k-guide.md
nupuruttarwar: yes, iecm and iavf are merged as part of idpf

Koushik-kopanati (approved):

docs/guides/security-guide.md
```diff
@@ -132,18 +119,25 @@ by the ca.crt (can copy the generated files from the server if client is not
...
 ### P4RT client
-
-The p4rt-ctl (P4RT client) will default to communicate via secure mode
-(port 9559). If certificates are not available, the P4RT client will attempt a
+The p4rt-ctl (P4RT client) will default to communicate via secure mode using
+port 9559. If certificates are not available, the P4RT client will attempt a
 connection using insecure client credentials as a fallback mechanism. Note that
-the communication may fail if infrap4d runs in secure mode only.
+the communication may fail if infrap4d runs in secure mode only. Both server and
+client should specify insecure mode for this to work.
```
ffoulkes, line 126: _should_ or _must_?

docs/guides/security-guide.md
```diff
@@ -132,18 +119,25 @@ by the ca.crt (can copy the generated files from the server if client is not
...
 
 ### gNMI client
 
-The gnmi-ctl or sgnmi_cli (gNMI clients) requests should be directed to port 9339.
-If infrap4d is enabled to run in insecure mode, a flag is available on client side
+The gnmi-ctl(gNMI client for DPDK) or sgnmi_cli (gNMI client for Intel IPU E2100) requests
```
ffoulkes, line 130: Unfortunately, the cool markup feature doesn't work (or I can't figure out how to make it work) on multiple lines.

"gnmi-ctl (the gNMI client for DPDK) and sgnmi_cli (the secure gNMI client) issue requests to port 9339. If infrap4d is started in insecure mode, a flag must be specified so the client also runs in insecure mode."

docs/guides/security-guide.md
```diff
@@ -113,11 +100,11 @@ More info on these values can be found on [this gRPC library documentation page]
...
-`grpc_open_insecure_ports` value to `true` to open insecure ports.
+`grpc_open_insecure_mode` value to `true` to allow insecure communication
 Also, make sure `certs` directory is removed from the default location mentioned
 above.
 
-To launch infrap4d with insecure ports 9339 and 9559 open:
+To launch infrap4d in insecure mode:
 
 ```bash
 $IPDK_RECIPE/install/sbin/infrap4d  -grpc_open_insecure_mode=true
```
ffoulkes, line 110: I would suggest defining an `IPDK_INSTALL` environment variable and changing the examples to, e.g., `$IPDK_INSTALL/sbin/infrap4d` instead of `$IPDK/install/sbin/infrap4d`.

The semantic component that gets replaced is the base install directory, not its parent.

This way, the command works regardless of whether the recipe is installed in `$IPDK/install` or `/opt/p4cp`.

scripts/security/setup_certs_tls_mode.sh
ffoulkes, line 1: The commit comment says only "Fix documentation with correct information." It says nothing about deleting the TLS mode setup script.

The comment should mention this, and it should explain _why_ the script is being deleted.

docs/guides/security-guide.md
nupuruttarwar: must

scripts/security/setup_certs_tls_mode.sh
nupuruttarwar: It's added as part of commit description - 

Remove scripts/security/setup_certs_tls_mode.sh for now as it doesn't
consider all the environments and format for infrap4d package and runtime
environment.

A revised script will be committed later to support all use cases

docs/guides/security-guide.md
nupuruttarwar: Sure, good idea. Will take it as a separate PR since it will touch multiple files

docs/guides/security-guide.md
```diff
@@ -132,18 +119,23 @@ by the ca.crt (can copy the generated files from the server if client is not
 on the same system as server).
 
 ### P4RT client
-
-The p4rt-ctl (P4RT client) will default to communicate via secure mode
-(port 9559). If certificates are not available, the P4RT client will attempt a
+The P4Runtime Control client will default to communicate via secure mode using
```
ffoulkes, line 122:
```suggestion
The P4Runtime Control client will default to communicating in secure mode using
```
A verb as a noun is a _gerund_, which calls for an _-ing_ ending.

"via secure mode" doesn't sound quite right to me. _via_ means 'through'. I think secure mode is something you're _in_, rather than something you _go through_.

docs/guides/security-guide.md
```diff
@@ -132,18 +119,23 @@ by the ca.crt (can copy the generated files from the server if client is not
...
+port 9559. If certificates are not available, the P4RT client will attempt a
 connection using insecure client credentials as a fallback mechanism. Note that
-the communication may fail if infrap4d runs in secure mode only.
+the communication will fail if infrap4d runs in secure mode only. Both server and
```
ffoulkes, line 125: Do we support a mode in which the _server_ falls back from secure mode to insecure mode? That's what "runs in secure mode _only_" implies. A more correct wording would be:

```suggestion
the communication will fail if only infrap4d runs in secure mode. Both server and
```

scripts/security/setup_certs_tls_mode.sh
ffoulkes: Excellent. But I'd still change the summary line. It's misleading.

ffoulkes (approved):
I've commented on grammar and punctuation in a couple of the revisions. Fairly minor stuff.

LGTM, though I'd be happier if you amended the summary line.

PR #168, opened 2023-06-15
Add SDE_INSTALL_DIR and DEPEND_INSTALL_DIR checks

ffoulkes:
- Generate fatal error if SDE_INSTALL_DIR is not defined.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -40,7 +40,9 @@ set(SDE_INSTALL_DIR "$ENV{SDE_INSTALL}" CACHE PATH
     "SDE install directory")
 
 if(SDE_INSTALL_DIR STREQUAL "")
-    message(FATAL_ERROR "SDE_INSTALL_DIR not defined!")
+    message(FATAL_ERROR "SDE_INSTALL_DIR (SDE_INSTALL) not defined!")
+elseif(DEPEND_INSTALL_DIR STREQUAL "")
+    message(FATAL_ERROR "DEPEND_INSTALL_DIR (DEPEND_INSTALL) not defined!")
 endif()
 
 set(HOST_DEPEND_DIR "" CACHE PATH "Host dependencies install directory")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -39,6 +39,10 @@ set(OVS_SOURCE_DIR "${CMAKE_SOURCE_DIR}/ovs/ovs" CACHE PATH
 set(SDE_INSTALL_DIR "$ENV{SDE_INSTALL}" CACHE PATH
     "SDE install directory")
 
+if(SDE_INSTALL_DIR STREQUAL "")
+    message(FATAL_ERROR "SDE_INSTALL_DIR not defined!")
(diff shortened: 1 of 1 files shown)


nupuruttarwar (commented):
change LGTM, should be checking other variables as well which are derived from env variables, say DEPEND_INSTALL?

ffoulkes:
> change LGTM, should we be checking other variables as well which are derived from env variables, say DEPEND_INSTALL?

Not necessarily. I just remembered why I didn't require that the various xxxx_INSTALL_DIR variables be defined.

The original premise was that all the packages would be installed to the same directories. I introduced support for separate directories, partly to make it easier to work on one package at a time during development, and partly to support Google's recommendation that gRPC and Protobuf _not_ be installed to system directories.

There are also a couple of use cases that don't necessarily require the full build infrastructure. It is currently possible to build and run the kernel monitor unit tests with nothing but target-syslibs. (Once we add switchlink_db unit testing, we will also need tommyds, which is part of target-utils.)

If we're not careful, we'll end up burdening simple tasks with the weight of the entire build. I'm not convinced that requiring something we don't actually need is better than erroring out with an undefined library.

nupuruttarwar:
I understand that part of krnlmon unit tests(switchlink module mainly) can built without SDE and also agree about not burdening simple tasks with the weight of the entire build.  But switchapi component in krnlmon accesses TDI API's, so we will need to have SDE libraries for unit testing switchapi module. We will add that support once we get there anyways.

My question about checking DEPEND_INSTALL was more about when are doing overall build using cmake command or make-all.sh.
make-all.sh checks if _SDE_DIR is non-empty but doesn't check _DEPS_DIR even though the build depends on both. Thoughts?

ffoulkes:
> switchapi component in krnlmon accesses TDI API's, so we will need to have SDE libraries for unit testing switchapi module. We will add that support once we get there anyways.

Code that accesses the actual hardware libraries is an integration test, not a unit test. Either we mock the SDE interface or we don't unit-test the code.

ffoulkes:
> make-all.sh checks if _SDE_DIR is non-empty but doesn't check _DEPS_DIR even though the build depends on both. Thoughts?

Stratum is no longer an optional component. I think it's safe to say that P4 Control Plane is inherently dependent on the Stratum dependencies.

It costs us little at build time to be dependent on the SDE and Stratum dependencies. These dependencies are built and installed independently of the P4 Control Build.

The real build-time cost is in the number of source files (especially C++ and Protobuf) we have to compile. The optimizations in the kernel monitor testing path come from our target selection and the number of P4 Control Plane source files the targets require. The SDE and Stratum dependencies don't figure into this.

There is therefore no real risk in requiring that DEPEND_INSTALL be defined.

And in the event that the original scenario occurs (SDE and Stratum dependencies share an install tree), we should be able to finesse the check by setting SDE_INSTALL and DEPEND_INSTALL to the same file path.

I will amend the PR to perform a second check.

ffoulkes:
Well-spotted and well-argued, by the way.

PR #171, opened 2023-06-16
Address issues in setup helper scripts

ffoulkes:
- In make-host-deps.sh, correct the definitions of the -B and -P options to specify that they take arguments.

- Remove -C alias for --cxx option. We don't need a short form.

- Change default number of jobs from 6 to 8, to be consistent with make-cross-deps.sh.

- Remove superfluous message.


nupuruttarwar (approved):
LGTM

5abeel (approved):
LGTM

PR #173, opened 2023-06-18
Format ovs-p4rt source code

ffoulkes:
- Use clang-format to ensure that ovs-p4rt source code is in canonical form.

- Change namespace from ovs_p4rt_cpp to ovs_p4rt.

- Move the bulk of ovs_p4rt.cpp into the ovs_p4rt namespace. The C interfaces remain outside the namespace.


nupuruttarwar (approved):
LGTM

I think it will be good to bring up the clang format with the team so that any new code is well formatted.

5abeel (approved):
LGTM

PR #174, opened 2023-06-20
Implement OVS Find-module

ffoulkes:
- Reimplement the OVS library interface logic as a cmake Find Module. In keeping with modern practice, define a target for each library.

- Change default value of the OVS_INSTALL_DIR variable from ${CMAKE_INSTALL_PREFIX} to $ENV{OVS_INSTALL}. The OVS install tree is an input to the P4 Control Plane build, not an output.

- Delete the OVS_SOURCE_DIR variable. The P4 Control Plane build uses the OVS install tree, not its source tree.

- Fix a couple of issues in the setup/prune.sh script.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -10,7 +10,7 @@ cmake_minimum_required(VERSION 3.15)
 # CMAKE_BUILD_TYPE must be set before the first project() command.
 set(CMAKE_BUILD_TYPE "RelWithDebInfo" CACHE STRING "Type of build to perform")
 
-project(networking-recipe VERSION 23.01 LANGUAGES C CXX)
+project(networking-recipe VERSION 23.07 LANGUAGES C CXX)
 
 list(APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake)
 
@@ -30,11 +30,8 @@ cmake_print_variables(CMAKE_STAGING_PREFIX)
 set(DEPEND_INSTALL_DIR "$ENV{DEPEND_INSTALL}" CACHE PATH
     "Dependencies install directory")
 
-set(OVS_INSTALL_DIR "${CMAKE_INSTALL_PREFIX}" CACHE PATH
+set(OVS_INSTALL_DIR "$ENV{OVS_INSTALL}" CACHE PATH
     "OVS install directory")
-
-set(OVS_SOURCE_DIR "${CMAKE_SOURCE_DIR}/ovs/ovs" CACHE PATH
-    "OVS source directory")
 
 set(SDE_INSTALL_DIR "$ENV{SDE_INSTALL}" CACHE PATH
     "SDE install directory")
@@ -178,6 +175,13 @@ elseif(ES2K_TARGET)
     include(es2k-driver)
 elseif(TOFINO_TARGET)
     include(tofino-driver)
+endif()
+
+if(WITH_OVSP4RT)
+    if(OVS_INSTALL_DIR STREQUAL "")
+        message(FATAL_ERROR "OVS_INSTALL_DIR not defined!")
+    endif()
+    find_package(OVS)
 endif()
 
 ########################
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -179,7 +179,7 @@ endif()
 
 if(WITH_OVSP4RT)
     if(OVS_INSTALL_DIR STREQUAL "")
-        message(FATAL_ERROR "OVS_INSTALL_DIR not defined!")
+        message(FATAL_ERROR "OVS_INSTALL_DIR (OVS_INSTALL) not defined!")
     endif()
     find_package(OVS)
 endif()
--- a/cmake/FindOVS.cmake
+++ b/cmake/FindOVS.cmake
@@ -8,7 +8,7 @@
 # Use pkg-config to search for the modules.
 #-----------------------------------------------------------------------
 find_package(PkgConfig)
-pkg_check_modules(PC_OVS QUIET libopenvswitch)
+pkg_check_modules(PC_OVS libopenvswitch)
 
 #-----------------------------------------------------------------------
 # Find libraries and include directories.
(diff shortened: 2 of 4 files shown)


cmake/FindOVS.cmake
```diff
@@ -0,0 +1,131 @@
...
+#-----------------------------------------------------------------------
+if(PC_OVS_VERSION)
+    set(OVS_VERSION ${PC_OVS_VERSION} CACHE STRING "OVS version")
+elseif(EXISTS "${OVS_INCLUDE_DIR}/openvswitch/version.h")
+    # Extract version string from version.h file.
+    file(STRINGS
+        "${OVS_INCLUDE_DIR}/openvswitch/version.h"
+        _version_string REGEX "OVS_PACKAGE_VERSION")
+    string(REGEX REPLACE
+        "[^0-9.]+([0-9.]+).*" "\\1" OVS_VERSION ${_version_string})
```
5abeel, line 62: What is this doing? Looking for a specific version for a pattern match?

CMakeLists.txt
```diff
@@ -30,12 +30,9 @@ cmake_print_variables(CMAKE_STAGING_PREFIX)
 set(DEPEND_INSTALL_DIR "$ENV{DEPEND_INSTALL}" CACHE PATH
     "Dependencies install directory")
 
-set(OVS_INSTALL_DIR "${CMAKE_INSTALL_PREFIX}" CACHE PATH
+set(OVS_INSTALL_DIR "$ENV{OVS_INSTALL}" CACHE PATH
```
nupuruttarwar, line 33: Is user expected to set OVS_INSTALL variable after this change?

CMakeLists.txt
```diff
@@ -180,6 +177,13 @@ elseif(TOFINO_TARGET)
     include(tofino-driver)
 endif()
 
+if(WITH_OVSP4RT)
+    if(OVS_INSTALL_DIR STREQUAL "")
+        message(FATAL_ERROR "OVS_INSTALL_DIR not defined!")
```
nupuruttarwar, line 182: Be consistent in messaging - message(FATAL_ERROR "OVS_INSTALL_DIR (OVS_INSTALL) not defined!")? Similar to line 40 and 42.

cmake/FindOVS.cmake
```diff
@@ -0,0 +1,131 @@
+# FindOVS.cmake - find Open vSwitch package.
+#
+# Copyright 2022-2023 Intel Corporation
```
nupuruttarwar, line 3: only 2023 since this is a new file introduced in 2023?

cmake/FindOVS.cmake
```diff
@@ -0,0 +1,131 @@
...
+#-----------------------------------------------------------------------
+find_package(PkgConfig)
+pkg_check_modules(PC_OVS QUIET libopenvswitch)
+
+#-----------------------------------------------------------------------
+# Find libraries and include directories.
+#-----------------------------------------------------------------------
+find_path(OVS_INCLUDE_DIR
+    NAMES "openvswitch/version.h"
+    PATHS ${PC_OVS_INCLUDE_DIRS}
```
nupuruttarwar, line 18: PC_OVS what does **C** stand for?

cmake/FindOVS.cmake
```diff
@@ -0,0 +1,131 @@
...
+    PATHS ${PC_OVS_LIBRARY_DIRS}
+)
+
+find_library(OVSDB_LIBRARY
+    names ovsdb
+    PATHS ${PC_OVS_LIBRARY_DIRS}
+)
+
+find_library(SFLOW_LIBRARY
+    names sflow
```
nupuruttarwar, line 37: Valid option according to cmake documentation - https://cmake.org/cmake/help/latest/command/find_library.html is uppercase **NAMES**
image

Also, I see lowercase **names** is not encoded in red indicating .cmake is not detecting it as valid option

cmake/FindOVS.cmake
```diff
@@ -0,0 +1,131 @@
...
+    string(REGEX REPLACE
+        "[^0-9.]+([0-9.]+).*" "\\1" OVS_VERSION ${_version_string})
+    set(OVS_VERSION ${PC_OVS_VERSION} CACHE STRING "OVS version")
+    unset(_version_string)
+endif()
+
+#-----------------------------------------------------------------------
+# Handle REQUIRED and QUIET arguments
+#-----------------------------------------------------------------------
+find_package_handle_standard_args(OVS
```
5abeel, line 70: Do we need a `REASON_FAILURE_MESSAGE` message here to indicate if one of the `REQUIRED_VARS` is missing? (not sure what the default messaging looks like, but should not fail silently/cryptic message)

CMakeLists.txt
nupuruttarwar: But make-all.sh script is setting the default for OVS_DIR to ovs/install if env var OVS_INSTALL is empty (https://github.com/ipdk-io/networking-recipe/blob/2dd83088a56ddd4fc66cdc3c22b127e42e947181/make-all.sh#L20). We don't do that in CMakeLists.txt. So if the user is building using cmake directly, they are expected to set OVS_INSTALL env variable?

nupuruttarwar (approved):
LGTM

cmake/FindOVS.cmake
ffoulkes: Good idea. I should give this a try.

ffoulkes: Ah. `REASON_FAILURE_MESSAGE` isn't supported until CMake 3.16. Our canonical minimum version is 3.15.

`FAIL_MESSAGE` (the related option) is "not recommended" (which doesn't stop eight of the standard find-modules in cmake-3.16 from using it).

ffoulkes: I did contrive to have `find_package_handle_standard_args()` fail in one of the tests I did during development. IIRC, its default error messages was somewhat cryptic.

5abeel: ok. Can look into this as future improvement if we want to merge this PR now

5abeel (approved):
LGTM

PR #176, opened 2023-06-21
Move cross-compile scripts to scripts/es2k

ffoulkes:
- Move `make-cross-ovs.sh` and `config-cross-recipe.sh` to the `scripts/es2k` directory. This helps declutter the root directory and makes it clear that these two scripts are specific to the ES2K target.
- At Sandeep's request, restore the scripts to the superproject (without removing them from scripts/es2k), to address a circular dependency in the CI build pipeline. (My preference would have been to use symlinks instead of duplicate copies, but there are some concerns about how git handles them.) The copies in the superproject will be removed once the pipeline change has taken effect.


ffoulkes:
This PR will require corresponding change(s) to the MEV build scripts.

n-sandeep:
Yes Derek, These changes will break eisting ACC builds. This needs a change in ACC build scripts as well.

n-sandeep:
can this PR wait?
We have a cyclic dependency here.

ACC build as part of OD will fail when we update manifest with SHA which has changed script location.
If we go ahead and update ACC to use a script from new file, since manifest is not updated with latest SHA even that will fail.
We have a solution here, where we can use the same branch name for both manifest and ACC script. We might have issues when updating this in release branch.

Is it possible to have a transition, 
we will have script in both locations, and once ACC build script is updated with new location then we can go ahead and remove the file from older directory

n-sandeep (approved):
LGTM. Older files will be removed once we update ACC scripts to pick these scripts from new location.

nupuruttarwar (approved):
LGTM

PR #177, opened 2023-06-21
Move cmake support files to setup/cmake directory

ffoulkes:
- Move files referenced by `setup/CMakeLists.txt` file to `setup/cmake` subdirectory. This should help reduce clutter in the parent directory.


5abeel (approved):
LGTM

setup/CMakeLists.txt
```diff
@@ -213,7 +213,7 @@ endif()
 # Protobuf compiler plugins to the relative paths of the library
 # directories.
 set(GRPC_INSTALL_RPATH $ORIGIN/../lib64:$ORIGIN/../lib)
-configure_file(grpc.patch.in ${CMAKE_SOURCE_DIR}/grpc.patch @ONLY)
+configure_file(cmake/grpc.patch.in ${CMAKE_SOURCE_DIR}/grpc.patch @ONLY)
```
n-sandeep, line 216: should the gRPC patch file also be part of the cmake directory?
As per internal discussions, i understand that going forward to fix any vulnerabilities in the dependent submodules we are going to add the patch which fixes the vulnerability. Should we create a patch directory?

ffoulkes: A lot of cmake projects treat their `configure_file()` input files as being associated with cmake. (They even give the files a `.cmake` suffix. I disagree with this practice -- `.cmake` should be reserved for included files that are written in the CMake command language.) I moved it as part of this PR because I'm trying to declutter the `setup` directory.

You're right, it would make sense to create a `patches` directory when we add the patch(es) for the gRPC vulnerabilities. I suppose we could move the `grpc.patch.in` to the directory at that time. I would probably rename the file at the same time, since we will then have more than one gRPC patch.

n-sandeep (approved):
Looks good to me

setup/CMakeLists.txt
ffoulkes: A follow-up note: it turns out that the CVE fixes affect a lot of files (~48+ if memory serves), which makes the notion of retrofitting just those patches a lot less attractive.

I'm currently experimenting with updating gRPC from v1.50.1 to v1.54.2 without changing anything else (https://github.com/ipdk-io/networking-recipe/pull/193). The host (x86 native) builds ran without error. The cross-compilation for ACC turned up a bug in `grpc/CMakeLists.txt` and some weaknesses in the way I'm patching the gRPC build.

The jury's still out on this one.

PR #184, opened 2023-06-25
Revise building-p4-programs.md

ffoulkes:


ffoulkes:
Practically speaking, this document is primarily about installing the ES2K P4 compiler and tools. This is a separate task from compiling the P4 program and deploying its artifacts. It's something you would do once, as apposed to the many times you would repeat the compile-prepare-deploy cycle once the tools were installed.

I would recommend making sections 1-3 as the basis of an "Installing" document and using sections 4-5 as the foundation of a separate "Building and Deploying" document.

PR #186, opened 2023-06-28
Revise security guide

nupuruttarwar:
Cleanup security guide and move instructions to install TLS certificate to separate guide

### DIFF
--- a/docs/guides/security-guide.md
+++ b/docs/guides/security-guide.md
@@ -19,16 +19,14 @@ The gRPC ports are secured using TLS certificates. A script and reference
 configuration files are available to assist in generating certificates and
 keys using OpenSSL. You may use other tools if you wish.
 
-The reference file uses a simple PKI where a self-signed key and certificate.
+The [reference files](https://github.com/ipdk-io/stratum-dev/tree/split-arch/tools/tls)
+uses a simple PKI where a self-signed key and certificate.
 The root level Certificate Authority (CA) is used to generate server-side
 key and cert files, and client-side key and cert files. This results in a
 1-depth level certificate chain, which will suffice for validation and
 confirmation but may not provide sufficient security for production systems.
 It is highly recommended to use well-known CAs, and generate certificates at
 multiple depth levels in order to conform to higher security standards.
-
-The reference files are available here:
-<https://github.com/stratum/stratum/tree/main/tools/tls>
 
 ### 2.1 Running in secure mode
 
--- a/docs/guides/security-guide.md
+++ b/docs/guides/security-guide.md
@@ -1,25 +1,19 @@
 # Security Guide for P4 Control Plane
 
-## Overview
+- [1. Overview](#1-overview)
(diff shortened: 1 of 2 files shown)


docs/guides/security-guide.md
```diff
@@ -1,141 +1,95 @@
...
--server_cert_file=[Server certificate file]
--server_key_file=[Server private key file]
-```
+#### 2.1.2 Requirements for gRPC clients
 
-For example:
+Under default conditions, the gRPC clients will require the TLS certificates
+to establish communication with infrap4d. The clients will need
+to use the same ca.crt file and the client.key and client.crt files signed
+by the ca.crt (can copy the generated files from the server if client is not
```
5abeel, line 52: This needs some editing for clarity.  Suggestion:
```
The clients will need to use the generated client.key and client.crt files signed
by the same CA (can copy the generated files from the server if client is not on the same system as server). 
```

docs/guides/security-guide.md
```diff
@@ -1,141 +1,95 @@
...
-    -server_key_file=/tmp/certs/stratum.key
-```
+- P4RT client
 
-### Client certificate verification
+    The P4Runtime Control client will default to communicating in secure mode
+using port 9559. If certificates are not available, the P4RT client will attempt
+a connection using insecure client credentials as a fallback mechanism.
+Note that the communication will fail if infrap4d runs in secure mode.
+Both server and client must specify insecure mode for this to work.
```
5abeel, line 61: Change `Both server and client must specify insecure mode`
to `Server miust specify insecure mode`

since client does not have a flag to specify insecure mode

docs/guides/security-guide.md
```diff
@@ -1,141 +1,95 @@
...
 
-## gRPC clients
+To launch clients in insecure mode:
 
-Under default conditions, the gRPC clients will require the TLS certificates
-to establish communication with infrap4d. The clients will need
-to use the same ca.crt file and the client.key and client.crt files signed
-by the ca.crt (can copy the generated files from the server if client is not
-on the same system as server).
+For DPDK target:
```
5abeel, line 83: Does this need to be broken down for different targets? We can simplify this by simply saying that user needs to pass the `-grpc_use_insecure_mode=true` flag

5abeel (changes requested):

docs/guides/security-guide.md
nupuruttarwar: I just realized that we have an option of insecure mode for gnmi client but not p4rt-ctl so we can't have a generic statement.

docs/guides/security-guide.md
nupuruttarwar: I specified different targets because we use sgnmi_cli for ES2K and gnmi-ctl for DPDK.

5abeel (approved):
LGTM

bharticemk (approved):

ffoulkes (approved):

PR #189, opened 2023-06-29
Add support for libunbound

ffoulkes:
- Implement Find module for libunbound package.

- Include Unbound library when linking ovs-vswitchd and ovs-testcontroller if package is installed.

### DIFF
--- a/cmake/FindUnbound.cmake
+++ b/cmake/FindUnbound.cmake
@@ -32,7 +32,7 @@ if(UNBOUND_VERSION)
 endif()
 
 #-----------------------------------------------------------------------
-# Handle REQUIRED and QUIETLY arguments
+# Handle REQUIRED and QUIET arguments
 #-----------------------------------------------------------------------
 find_package_handle_standard_args(UNBOUND
     REQUIRED_VARS
--- a/cmake/FindUnbound.cmake
+++ b/cmake/FindUnbound.cmake
@@ -1,0 +1,56 @@
+# FindUnbound.cmake - Find libunbound package.
+#
+# Copyright 2023 Intel Corporation
+# SPDX-License-Identifier: Apache 2.0
+#
+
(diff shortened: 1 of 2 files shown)


cmake/FindUnbound.cmake
```diff
@@ -0,0 +1,56 @@
...
+    VERSION_VAR
+        UNBOUND_VERSION
+)
+
+mark_as_advanced(UNBOUND_INCLUDE_DIR UNBOUND_LIBRARY UNBOUND_VERSION)
+
+#-----------------------------------------------------------------------
+# Define library target
+#-----------------------------------------------------------------------
+add_library(unbound UNKNOWN IMPORTED)
```
n-sandeep, line 50: should we add in our document that libunbound should be manually installed?
or is it part of legacy OVS doc?

n-sandeep (approved):
Minor comment, LGTM

5abeel (approved):
LGTM

cmake/FindUnbound.cmake
ffoulkes: It's really a function of what features of OVS the customer wants to use or support. We're just adapting to it by providing access to the library when it exists.

We've been running without `libunbound` for over a year. It only became an issue when Sabeel built P4 Control Plane on a system that actually had it installed. The OVS build detected the installation and enabled conditional code to use it, which broke our build because we didn't have support for the library in our build system.

For the time being, I think we can defer to the OVS documentation, but you raise an excellent point. (hat tip.)  To serve the needs of our customers, we should have a solid understanding of the requirements (both mandatory and optional) of all the packages we supply or support; we should provide for them in our build systems; and we should consider them in our setup procedures, validation procedures, and documentation.

satish153 (approved):
LGTM

nupuruttarwar (approved):
LGTM

PR #192, opened 2023-07-01
Building Stratum Dependencies document

ffoulkes:
Generic instructions for building the Stratum dependencies using CMake.
ES2K-specific instructions (using the helper scripts) will follow.

### DIFF
--- a/setup/building-stratum-deps.md
+++ b/setup/building-stratum-deps.md
@@ -93,11 +93,13 @@ The CMake build script supports the following configuration options.
 | `DOWNLOAD` | Boolean | Whether to download the source repositories. (Default: TRUE)
 | `FORCE_PATCH` | Boolean | Whether to specify the force (`-f`) option when patching a downloaded repository. (Default: FALSE) |
 | `ON_DEMAND` | Boolean | Whether to build only the specified target(s). If this option is FALSE, all targets will be built. (Default: FALSE) |
-| `USE_LDCONFIG` | Boolean | Whether to use `ldconfig` to update the loader cache after installing a module. Only valid if `USE_SUDO` is enabled. (Default: FALSE) |
+| `USE_LDCONFIG` | Boolean | Whether to use `ldconfig` to update the loader cache[1] after installing a module. Only valid if `USE_SUDO` is enabled. (Default: FALSE) |
 | `USE_SUDO` | Boolean | Whether to use `sudo` to install each module. (Default: FALSE) |
 
 Boolean values are (`TRUE`, `YES`, `ON`) and (`FALSE`, `NO`, `OFF`).
 They may be upper or lower case.
+
+[1] See the `ldconfig` man page for more information.
 
 ## 6 Examples
 
--- a/setup/building-stratum-deps.md
+++ b/setup/building-stratum-deps.md
@@ -72,8 +72,8 @@ need to log in as `root` or run from an account that has `sudo` privilege.
 By default, the cmake build script downloads the source code for each module
 and then builds and installs it.
 
-If you are working from a source distribution, or if the source files have
-were downloaded by a previous build, you can override the download stage by
+If you are working from a source distribution, or if the source files were
+downloaded by a previous build, you can override the download stage by
 specifying:
 
 ```cmake
--- a/setup/building-stratum-deps.md
+++ b/setup/building-stratum-deps.md
@@ -6,22 +6,22 @@ be built and installed on the system.
 
 This document explains how to build and install the Stratum dependencies.
 
-> Note: For the Intel IPU E2100, see
+> Note: For the Intel&reg; IPU E2100, see
 [Building Stratum Dependencies for ES2K](building-stratum-deps-es2k.md).
 
 ## 1 Prerequisites
 
 ### 1.1 CMake
 
-`cmake 3.15` or above must be installed.
+CMake 3.15 or above must be installed.
 
-Avoid versions 3.24.x and 3.25.x. They have an issue that causes the Protobuf
-build to fail.
+Avoid versions 3.24 and 3.25. There is an issue in cmake that causes the
+Protobuf build to fail. This problem was fixed in version 3.26.
 
 ### 1.2 OpenSSL
 
-`OpenSSL 1.1` must be installed. The Stratum dependencies use OpenSSL instead
-of building and installing BoringSSL.
+OpenSSL 1.1 must be installed. P4 Control Plane uses OpenSSL instead of
+BoringSSL.
 
 ## 2 Source Code
 
@@ -103,7 +103,7 @@ They may be upper or lower case.
 
 ### 6.1 Non-root build
 
-To build and install as a nonprivileged user:
+To build and install as a non-privileged user:
 
 ```bash
 cd setup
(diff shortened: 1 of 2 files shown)


setup/building-stratum-deps.md
```diff
@@ -0,0 +1,183 @@
+# Building the Stratum Dependencies
+
+Stratum is the component of `infrap4d` that implements the P4Runtime and gNMI
+(OpenConfig) services. It requires that a number of third-party libraries
+be built and installed on the system.
+
+This document explains how to build and install the Stratum dependencies.
+
+> Note: For the Intel IPU E2100, see
+[Building Stratum Dependencies for ES2K](building-stratum-deps-es2k.md).
```
5abeel, line 10: Self reference?

5abeel: On that note, I think it makes more sense to have the document in `docs` folder rather than in `setup`. If looking for any build instructions, users will generally tend to look in docs

setup/building-stratum-deps.md
```diff
@@ -0,0 +1,183 @@
...
+  -DDOWNLOAD=no -DFORCE_PATCH=yes
+```  
+
+when you configure the cmake build.
+
+## 5 Build Options
+
+The CMake build script supports the following configuration options.
+
+| Option | Type | Description |
```
5abeel, line 89: The table looks weird in GitHub review. please verify that when rendered, it looks ok

setup/building-stratum-deps.md
```diff
@@ -0,0 +1,183 @@
...
+
+The source files will be downloaded, built, and installed in the `setup/install`
+directory. The targets will be built in parallel, using eight threads.
+
+CMake will generate all of its temporary files in the `setup/build` directory.
+This is called an "out-of-source" build.
+
+### 6.2 Non-root build to system directory
+
+To build as a nonprivileged user and use `sudo` to install the libraries to
```
5abeel, line 122: `non-privileged`? (with hyphen)

setup/building-stratum-deps.md
```diff
@@ -0,0 +1,183 @@
...
+### 6.4 On-demand build
+
+To build and install just gRPC and its dependencies:
+
+```bash
+cmake -B build -DON_DEMAND=yes -DCMAKE_INSTALL_PREFIX=~/hostdeps
+cmake --build build -j6 --target grpc
+```
+
+The repositories for grpc and its dependencies will be downloaded, built, and
```
5abeel, line 160: `gRPC` instead of `grpc`

setup/building-stratum-deps.md
ffoulkes: > Self reference?

No, it's a reference to an ES2K-specific document that I'm working on.

I mentioned this in the comment at the top of this page.

> Generic instructions for building the Stratum dependencies using CMake.
ES2K-specific instructions (using the helper scripts) will follow.

ffoulkes: > On that note, I think it makes more sense to have the document in `docs` folder rather than in `setup`. If looking for any build instructions, users will generally tend to look in docs

I feel very strongly that the instructions for building the dependencies belong with the dependencies. Building and installing the dependencies is a precursor to building and installing P4 Control Plane, and is no more a part of P4 Control Plane than the SDE is.

In fact, I'm proposing to spin the Stratum Dependencies off as an independent package with its own repository, and refer to it rather than including it in the networking recipe.

See https://wiki.ith.intel.com/display/ndvswitch/P4CP+-+Stratum+Dependencies for more information.

setup/building-stratum-deps.md
ffoulkes: It looks pretty good. To see for yourself, click the ellipsis (. . .) in the upper right corner of the file panel (to the right of the "Comment on this" icon, below the "Review Changes" button) and select "View File". GitHub will give you preview of the page.

setup/building-stratum-deps.md
ffoulkes: Interesting. A web search turns up `nonprivileged`, `non-privileged`, and `unprivileged`.

1. Looking at the first two side-by-side, I find `non-privileged` somewhat easier to read than `nonprivileged`, which is reason enough to prefer it.
2. One of the first hits I got brought me to `unprivileged`, suggesting that the unhyphenated form is less common or less correct.
3. `unprivileged` might be more correct, but has a pejorative connotation that I'd like to avoid.

`non-privileged` it is.

setup/building-stratum-deps.md
ffoulkes: Actually I wound up rewording the first part of the sentence. It was essentially the same as line 153. That kind of repetition makes for a boring document.

setup/building-stratum-deps.md
5abeel: hmmm I see. I would atleast like to see a direct link from the "main build" document in `docs` folder to this then.
We can revisit once all documentation is in place. Approving this PR.

5abeel (approved):
LGTM

setup/building-stratum-deps.md
ffoulkes: The plan is to turn `es2k-guide.md` into a list of steps to take, with each step linking to the document for that step.

setup/building-stratum-deps.md
```diff
@@ -0,0 +1,185 @@
...
+
+If this is case, use the appropriate command to unpack the source files.
+
+## 3 Install Location
+
+You will need to decide where on your system you would like to install the
+dependencies. This location (the "install prefix") must be specified when you
+configure the build.
+
+It is recommended that you _not_ install the dependency libraries in `/usr` or
```
nupuruttarwar, line 60: do _not_?

setup/building-stratum-deps.md
```diff
@@ -0,0 +1,185 @@
...
+If you plan to install the dependency libraries in a system directory, you will
+need to log in as `root` or run from an account that has `sudo` privilege.
+
+## 4 Downloaded Source vs. Source Package
+
+By default, the cmake build script downloads the source code for each module
+and then builds and installs it.
+
+If you are working from a source distribution, or if the source files have
+were downloaded by a previous build, you can override the download stage by
```
nupuruttarwar, line 76: "or if the source **files have were downloaded** by a previous build" 
Wording seems off. Please check once. May be "if the source files **you have** were download"

nupuruttarwar (approved):
Minor comments. LGTM

setup/building-stratum-deps.md
ffoulkes: I knew this form from use, but I had to look it up to provide an explanation.

_It is recommended that you not ..._ is subjunctive. It is correct. One of the sources I browsed says it is considered dated or formal.

_It is recommended that you do not ..._ is indicative. It is also correct. One of the sources I browsed says that "do" is redundant when used with "recommended".

My personal inclination is use _do_ when using the indicative form ("Google recommends that you do not") and omit it when using the subjunctive form ("It is recommended that you not").

I prefer the subjunctive mood in this case. It's stronger than "we suggest" and weaker than "you should", and it seems more appropriate when it's not accompanied by an explanation.

setup/building-stratum-deps.md
ffoulkes: I think was changing "if the source files have already been downloaded" to "if the source files were downloaded by a previous build" and missed deleting the "have".

Good catch. Thank you.

PR #194, opened 2023-07-04
Fix force-patch omission in make-cross-deps.sh

ffoulkes:
- Supply missing definition of -f short option.


bharticemk (approved):
LGTM

nupuruttarwar (approved):
LGTM

PR #206, opened 2023-07-19
Add list of Stratum ES2K unit tests

ffoulkes:
Specifies the Stratum tests to be run by CI.

Not used yet (but soon, I hope).

This PR depends on https://github.com/ipdk-io/stratum-dev/pull/90 and a stratum submodule update.


.github/es2k-tests.txt
```diff
@@ -0,0 +1,8 @@
+//stratum/glue/...
+//stratum/hal/lib/p4/...
+//stratum/lib/...
+//stratum/public/...
+//stratum/hal/lib/tdi:all
+//stratum/hal/lib/tdi/es2k:es2k_chassis_manager_test
```
5abeel, line 6: Im not too familiar with bazel, but why do we need to specify individual files/testcases if line 5 already has `all`?

ffoulkes: `//stratum/hal/lib/tdi:all` selects just the tests in `//stratum/hal/lib/tdi`.

IIRC, `//stratum/hal/lib/tdi/...` would select everything in TDI and below. We don't want to do that, though, because it would pick up the es2k and tofino tests as well (with the wrong --define target=<target> and SDE).

nupuruttarwar (approved):
LGTM

aashishkuma (commented):
tests under "//stratum/hal/lib/tdi" are target-agnostic common tests. Can we have "tdi-common-tests.txt" file which should contain all the tests that are target-agnostic? And remove the references of "//stratum/hal/lib/tdi:all" from both es2k and dpdk test files and keep only target specific tests in their respective file? just to bring some uniformity in the code structure. 

so if somebody wants to run for example dpdk-tests, they can run "tdi-common" and "dpdk-test". later if other targets are introduced, the same common test file can be referenced and will avoid code duplication, else you will be keep on appending the same common tests in all the target specific test files.

5abeel (approved):
LGTM

PR #208, opened 2023-07-20
Add check for missing grpc cpp plugin

ffoulkes:
- It turns out that find_program() gained support for the REQUIRED parameter in cmake 3.18. Add an explicit check to ensure that the build fails if we are unable to find grpc_cpp_plugin in an earlier version of cmake.


5abeel (approved):
LGTM

nupuruttarwar (approved):
LGTM

PR #209, opened 2023-07-21
Building dependencies for the ES2K ACC

ffoulkes:
How to build the Stratum dependencies for cross-compilation to the ARM Compute Complex (ACC) of the Intel&reg; IPU E2100.

### DIFF
--- a/setup/building-es2k-stratum-deps.md
+++ b/setup/building-es2k-stratum-deps.md
@@ -31,19 +31,19 @@ between client applications and `infrap4d`.
 There are several things you need to take care of on your development
 system before you can build.
 
-### CMake
+### 2.1 CMake
 
 CMake 3.15 or above must be installed.
 
 Avoid versions 3.24 and 3.25. There is an issue in cmake that causes the
 Protobuf build to fail. The issue was fixed in version 3.26.
 
-### OpenSSL
+### 2.2 OpenSSL
 
 OpenSSL 1.1 must be installed. P4 Control Plane uses OpenSSL instead of
 BoringSSL.
 
-### ACC SDK
+### 2.3 ACC SDK
 
 Install the ACC SDK on your system. See
 [Installing the ACC SDK](../docs/guides/es2k/installing-acc-sdk.md) for
@@ -68,7 +68,7 @@ The build script for the Stratum dependencies is in the `setup` directory.
 
 ## 4. Host Dependencies
 
-### Host install location
+### 4.1 Host install location
 
 You will need to decide where on your system you would like to install the
 Host dependencies. This location (the "install prefix") must be specified
@@ -88,7 +88,7 @@ need to log in as `root` or run from an account that has `sudo` privilege.
 The distribution includes a helper script (`make-host-deps.sh`) that can be
 used to build the Host dependencies.
 
-### Example: non-root build
+### 4.2 Example: non-root build
 
 To install the dependencies in a user directory:
 
@@ -99,7 +99,7 @@ To install the dependencies in a user directory:
 The source files will be download and built, and the results of the build
 will be installed in `/home/peabody/hostdeps`.
 
-### Example: root build
+### 4.3 Example: root build
 
 ```bash
 ./make-host-deps.sh --prefix=/opt/p4cp/x86deps --sudo
@@ -107,7 +107,7 @@ will be installed in `/home/peabody/hostdeps`.
 
 The `--sudo` parameter may be omitted if you are logged in as `root`.
 
-### Example: complete build
+### 4.4 Example: complete build
 
 By default, the helper script does a *minimal* build, consisting only of
 the components that are needed to support cross-compilation. Specify the
@@ -117,7 +117,7 @@ the components that are needed to support cross-compilation. Specify the
 ./make-host-deps.sh --prefix=$HOME/hostdeps --full
 ```
 
-### make-host-deps.sh
+### 4.5 make-host-deps.sh
 
 The helper script supports the following options:
 
@@ -140,23 +140,23 @@ Options:
   --sudo              Use sudo when installing (Default: false)
 ```
 
-## 4. Target Dependencies
+## 5. Target Dependencies
 
-### Target install location
-
-under construction
-
-### Cross-compilation
+### 5.1 Target install location
 
 under construction
 
-### Example: sysroot build
+### 5.2 Cross-compilation
+
+under construction
+
+### 5.3 Example: sysroot build
 
 ```bash
 ./make-cross-deps.sh --host=/opt/x86deps --prefix=//opt/p4cp/deps
 ```
 
-### make-cross-deps.sh
+### 5.4 make-cross-deps.sh
 
 The helper script supports the following options:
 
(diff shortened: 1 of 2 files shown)


setup/building-es2k-stratum-deps.md
```diff
@@ -0,0 +1,221 @@
...
+```
+
+You may omit the `--recursive` option if you are only interested in building
+the dependencies.
+
+You may substitute your own local directory name for `ipdk.recipe`.
+
+The build script for the Stratum dependencies is in the `setup` directory.
+
+TThe source code for the dependencies is not part of the distribution.
```
n-sandeep, line 60: typo here, there is an extra 'T' in The.

setup/building-es2k-stratum-deps.md
```diff
@@ -0,0 +1,221 @@
...
+Remove the `build` directory from the previous build.
+
+```bash
+rm -fr build
+```
+
+Now run the build script:
+
+```bash
+./make-cross-deps.sh --host=HOSTDEPS --prefix=PREFIX
```
n-sandeep, line 212: Maybe, i am missing some context here.
"make-cross-deps.sh" script only need '--prefix' and '--toolchain' parameters right?

Also, will the grpc patch would have applied while building HOST, will it also work fine when we run this script?

ffoulkes: `make-cross-deps.sh` requires host dependencies (we use the native protobuf compiler to compile .proto to .cc and .h), install prefix, and toolchain (which is normally provided by the CMAKE_TOOLCHAIN_FILE environment variable). `--toolchain` exists primarily to allow you to override the environment variable.

I've been having problems with the --force parameter since we updated grpc. When I was testing out command sequences, I discovered that I could delete the `build` directory and do another build, and it skipped the download and patch steps, so that's what I described in the document.

ffoulkes: This behavior appears to be related to specifying a full Git SHA instead of a branch or tag when you download.

ffoulkes: I just tried this again. It's re-downloading grpc (because one of the files has been patched and therefore doesn't match an expected signature?), but it apparently doesn't perform the patch step. But that's okay, because the patch only applies to the host build (something that hadn't occurred to me before)!

setup/building-es2k-stratum-deps.md
```diff
@@ -0,0 +1,221 @@
...
+
+## 2. Preparing the System
+
+There are several things to do before you can build the dependencies.
+
+- Install CMake 3.15 or above
+
+  Avoid versions 3.24 and 3.25. They cause the dependencies build to fail.
+
+- Install OpenSSL 1.1
```
5abeel, line 33: We have OpenSSL version 3.0 in some versions of Linux such as Fedora 37, Rocky Linux 9.1.

setup/building-es2k-stratum-deps.md
```diff
@@ -0,0 +1,221 @@
...
+In order to cross-compile for the ACC, you will need to define a number
+of environment variables. This is typically done by putting the bash
+commands in a file (e.g. `es2k-setup.env`) and using the `source` command
+to execute it. We recommend removing execute permission from the file
+(`chmod a-x setup.env`) to remind yourself to source it, not run it.
+
+For example:
+
+```bash
+ACC_SDK=<acc-sdk-directory>
```
5abeel, line 126: Is this a pre-defined environment variable that we use? Im wondering if there might be confusion with the _IPU SDK_ that users will be using going forward

setup/building-es2k-stratum-deps.md
ffoulkes: To the best of my knowledge, we have only ever tested against OpenSSL 1.1.

You and I discussed this a couple of months ago. Did you every get a definitive answer regarding building and testing against, and compatibility with, OpenSSL 3.0?

Also, as I recall, there was also an argument that we should move to 3.1.1. I don't remember what it was: possibly an IPAS recommendation, or the availability of something FIPS-related?

In any case, I'd like some assurance before saying anything about 3.0 in the documentation.

setup/building-es2k-stratum-deps.md
ffoulkes: ACC_SDK is local to this script, as are P4CPBASE, AARCH64, and SYSROOT.

ACC_SDK and P4CPBASE are discrete variables that the user sets directly. I do this so they don't have to edit multiple occurrences or change something that is part of a longer string. If this were a template file, the values would be marked for replacement (@ACC_SDK@ and @P4CPBASE@) at configuration time.

AARCH64 and SYSROOT are local variables that represent the two directory trees we are mapping.

None of these variables is `export`ed. I debated making them explicitly `local` and possibly `unset`ting them at the end of the script, but that seemed a tad excessive.

The variable that we are publishing are the ones documented in lines 154-165.

setup/building-es2k-stratum-deps.md
5abeel: yes, I remember that discussion.
I believe most of the testing has been performed on Fedora 37 and Rocky Linux 9.1 which has OpenSSL 3.0 by default. 

Let's re-confirm before adding anything to documentation. I am ok with this for now for initial doc.

setup/building-es2k-stratum-deps.md
ffoulkes: I've gone through several iterations of this. First, I added a bunch of comments to the sample configuration script. This seemed excessive, especially for ten lines of bash in a document, so I removed the comments and added 28 lines explanatory text the document. I've just push a change that adds three comments and two blank lines to the document, and moves one of the existing lines. I think this will do for now.

I intend to extract this section and make it a separate document (it applies to cross-compiling P4CP as well), but I decide it was better to make the documentation available in this form, right now, instead of waiting for the P4CP document to be complete.

setup/building-es2k-stratum-deps.md
```diff
@@ -0,0 +1,226 @@
...
+It is downloaded by the build script.
+
+## 4. Building the Host Dependencies
+
+First, decide where to install the Host dependencies. This location (the
+"install prefix") must be specified when you configure the build.
+
+It is recommended that you *not* install the Host dependencies in `/usr` or
+`/usr/local`. It will be easier to rebuild or update the dependencies if
+their libraries are mingled with other libraries.
```
nupuruttarwar, line 70: libraries are **not** mingled?

nupuruttarwar (approved):
Minor comment, LGTM

setup/building-es2k-stratum-deps.md
ffoulkes: Pfft. You're right.

setup/building-es2k-stratum-deps.md
```diff
@@ -67,7 +67,7 @@ First, decide where to install the Host dependencies. This location (the
 
 It is recommended that you *not* install the Host dependencies in `/usr` or
 `/usr/local`. It will be easier to rebuild or update the dependencies if
-their libraries are mingled with other libraries.
+their libraries not are mingled with other libraries.
```
nupuruttarwar, line 70: libraries are not mingled?

ffoulkes: :headdesk:

nupuruttarwar (approved):
LGTM

5abeel (approved):
LGTM

setup/building-es2k-stratum-deps.md
ffoulkes: I've created a Jira task to address this issue: https://jira.devtools.intel.com/browse/NVFMSA-8984

PR #212, opened 2023-07-25
Minor edits to gnmi-ctl.rst

ffoulkes:


docs/clients/gnmi-ctl.rst
```diff
@@ -287,7 +287,7 @@ Limitations/Note
     ``PSA architecture`` P4 compiler uses and emits the PIPELINE name defined by the user in the program.
                          Currently it supports only ingress pipelines.
```
nupuruttarwar, line 288: ingress pipeline.

nupuruttarwar (approved):
Minor comment, LGTM

docs/clients/gnmi-ctl.rst
ffoulkes: The plural is actually correct in this case. _Ingress pipelines_ is an abstract noun, referring to a class of pipelines. Abstract nouns are usually singular.

If we were to change it _pipeline_, we'd have to add a definite article (_a_, _an_, or _the_), which doesn't make sense, because nothing in the context tells us which ingress pipeline we're referring to.

Crazy, no?

nupuruttarwar: I added a comment to specify that at a time we support setting only 1 forwarding ingress pipeline. The word pipelines gives an impression that you can set multiple pipelines at once.

You are right, we should say forwarding ingress pipeline

5abeel (approved):
LGTM

ffoulkes:
I'm going to go ahead and merge this in its present form, for the sake of the corrections it makes. We can follow up on the correct rewording of "Currently it supports only ingress pipelines" (an existing phrase) once we've sorted out the ambiguities in the phrase.

PR #216, opened 2023-07-26
Building P4 Control Plane for ES2K ACC

ffoulkes:
How to build P4 Control Plane for the ARM Compute Complex (ACC) of the Intel&reg; IPU E2100.

### DIFF
--- a/docs/guides/es2k/building-p4cp-acc.md
+++ b/docs/guides/es2k/building-p4cp-acc.md
@@ -71,7 +71,7 @@ To build OVS and install it in the sysroot directory under `/opt/ipdk/ovs`:
 
 Options:
 
-* `--prefix=PREFIX` - where to install OVS
+- `--prefix=PREFIX` - where to install OVS
 
 The `//` at the beginning of the prefix path is a shortcut provided by
 the helper script. It will be replaced with the sysroot directory path.
--- a/docs/guides/es2k/building-p4cp-acc.md
+++ b/docs/guides/es2k/building-p4cp-acc.md
@@ -29,7 +29,7 @@ setup file, in order to cross-compile P4 Control Plane for the ACC.
 
 ## Fetch Source Code
 
-If you have not already done so, you will to get a copy of the IPDK
+If you have not already done so, you will need to get a copy of the IPDK
 networking-recipe repository and its submodules:
 
 ```bash
@@ -40,15 +40,28 @@ You may substitute your own local directory name for `ipdk.recipe`.
 
 ## Build with OVS
 
+### Setup
+
+First, change to the source directory:
+
+```bash
+cd ipdk.recipe
+```
+
+If you have not already done so, source the file that the defines the
+[target build environment variables](../../../setup/building-es2k-stratum-deps.md#5-defining-the-target-build-environment).
+For example:
+
+```bash
+source setup/es2k-setup.env
+```
+
 ### Building OVS
 
 The distribution includes a helper script (`make-cross-ovs.sh`) that can be
 used to build OVS for P4 Control Plane.
 
-- The `--help` (`-h`) option lists the parameters the helper script supports
-
-- The `--dry-run` (`-n`) option displays the parameter values without
-  running CMake
+> The `--help` (`-h`) option lists the parameters the helper script supports.
 
 To build OVS and install it in the sysroot directory under `/opt/ipdk/ovs`:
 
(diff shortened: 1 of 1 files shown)


docs/guides/es2k/building-p4cp-acc.md
```diff
@@ -0,0 +1,132 @@
...
+
+The options are:
+
+- `--host=HOST` - path to the Host dependencies
+- `--deps=DEPS` - path to the Target dependencies
+- `--ovs=OVS` - path to the OVS installation
+- `--sde=SDE` - path to the SDE installation
+- `--prefix=PREFIX` - where to install P4 control plane
+
+The `//` at the beginning of the prefix path is a shortcut provided by
```
5abeel, line 104: Duplicate of line 76-77, can be removed

docs/guides/es2k/building-p4cp-acc.md
```diff
@@ -0,0 +1,132 @@
...
+
+## Install Stratum Dependencies
+
+See [Building Stratum Dependencies for the ACC](../../../setup/building-es2k-stratum-deps.md)
+for instructions.
+
+You will need the Host depencies, Target dependencies, and Build environment
+setup file, in order to cross-compile P4 Control Plane for the ACC.
+
+## Fetch Source Code
```
5abeel, line 30:
```suggestion
## Fetch the Source Code
```

docs/guides/es2k/building-p4cp-acc.md
```diff
@@ -0,0 +1,132 @@
...
+```bash
+cd ipdk.recipe
+```
+
+If you have not already done so, source the file that the defines the
+[target build environment variables](../../../setup/building-es2k-stratum-deps.md#5-defining-the-target-build-environment).
+For example:
+
+```bash
+source setup/es2k-setup.env
```
n-sandeep, line 56: This example is misleading, user might think this is the command and reports there is no file named "es2k-setup.env"

docs/guides/es2k/building-p4cp-acc.md
```diff
@@ -0,0 +1,132 @@
...
+    --prefix=//opt/ipdk/p4cp
+```
+
+The `--no-ovs` option excludes OVS support.
+
+Now perform the build:
+
+```bash
+cmake --build build -j8 --target install
+```
```
5abeel, line 132: Should we add a cleanup section?

docs/guides/es2k/building-p4cp-acc.md
n-sandeep, line 30: As per the document flow, is it better to 
"Fetch Source Code" section and then "Install Stratum Dependencies" ?

docs/guides/es2k/building-p4cp-acc.md
```diff
@@ -0,0 +1,132 @@
...
+
+### Building P4 Control Plane
+
+Now use CMake to build and install P4 Control Plane:
+
+```bash
+cmake --build build -j8 --target install
+```
+
+## Build without OVS
```
n-sandeep, line 115: If user doesn't want to build with OVS, then he skips "Build with OVS" and jumps to section "Build without OVS" which doesnt have "Setup" sub-section

docs/guides/es2k/building-p4cp-acc.md
ffoulkes: I'll try expanding the explanation in the lead-in paragraph.

docs/guides/es2k/building-p4cp-acc.md
ffoulkes: I've added instructions for doing a clean build.

5abeel (approved):
LGTM

docs/guides/es2k/building-p4cp-acc.md
ffoulkes: I wound up changing the example to "command with placeholder", followed by an explanation.

docs/guides/es2k/building-p4cp-acc.md
ffoulkes: I'm treating "Install Stratum Dependencies" as a precursor step. I was originally going to put it in "Prepare the System", but decided that it should be given a little more emphasis.

The plan is to move the Stratum dependencies to a separate repository (not a submodule). I'm kind of paving the way for that.

docs/guides/es2k/building-p4cp-acc.md
```diff
@@ -0,0 +1,152 @@
...
+## Define the Environment
+
+First, change to the source directory:
+
+```bash
+cd ipdk.recipe
+```
+
+Source the file that defines the
+[target build environment variables](../../../setup/building-es2k-stratum-deps.md#5-defining-the-target-build-environment).
```
nupuruttarwar, line 50: Suggestion: Instead of doing a relative path, can we add the link to document and section directly. This way our documents are relocatable.

nupuruttarwar (approved):
LGTM

docs/guides/es2k/building-p4cp-acc.md
ffoulkes: What do you mean by "adding the link to the document and section directly?"

nupuruttarwar: I meant something like this `https://github.com/ipdk-io/networking-recipe/blob/main/setup/building-es2k-stratum-deps.md#5-defining-the-target-build-environment`

ffoulkes: Ah: full URL. Relative links make development easier, especially when the documents you're working on haven't been committed. I figured I'd change to URLs as part of moving the Stratum dependencies to their own repository (at which point they will become necessary).

nupuruttarwar: That sounds good too. I am thinking of suggesting that we make these documents part of IPU SDK. We can scrub the documents for relative links if and when we do.

PR #217, opened 2023-07-27
Remove Customized Lookup table references

n-sandeep:
This also includes, removing p4runtime wheel file in P4CP repo
Earlier the implementation for Lookup table was under extern type with new table type, and to achieve this we had customized p4runtime.proto and p4info.proto. For usecases which needed python p4runtime client, we have to provide a customized python wheel file to have these proto changes.
Now, based on community recommendation, we have used regular table entry instead of extern. Removed customized p4runtime.proto and p4info.proto changes. Going forward we dont need LUT definitions in p4rt-ctl and also customized python wheel file.


nupuruttarwar:
Can we add an explanation that why this is being removed and mention that now we are using add-entry command instead of add-lut-entry for IPv4 LUT. 
whl file is removed because we support opensource p4runtime=1.3.0?

ffoulkes:
Does this require a change in the P4Runtime submodule reference?

n-sandeep:
> Can we add an explanation that why this is being removed and mention that now we are using add-entry command instead of add-lut-entry for IPv4 LUT. whl file is removed because we support opensource p4runtime=1.3.0?

Explain where? in the PR commit message?
Yes, we will be using regular pip3 to install p4runtime.

n-sandeep:
> > Can we add an explanation that why this is being removed and mention that now we are using add-entry command instead of add-lut-entry for IPv4 LUT. whl file is removed because we support opensource p4runtime=1.3.0?
> 
> Explain where? in the PR commit message? Yes, we will be using regular pip3 to install p4runtime.

Done, updated commit message

ffoulkes (commented):
The p4runtime wheel is only one part of the equation.

Stratum and its clients use C++ code generated from contents of the p4runtime submodule.

Does the p4runtime submodule reference need to be updated?
If not, why not?

n-sandeep:
> The p4runtime wheel is only one part of the equation.
> 
> Stratum and its clients use C++ code generated from the contents of the p4runtime submodule.
> 
> Does the p4runtime submodule reference need to be updated? If not, why not?

We have already removed customized p4runtime code from stratum (removed LUT manager) and also update p4runtime submodule to not use customized code instead a latest merge from opensource p4lang/P4runtime.
This PR was to cleanup some code which was never removed

nupuruttarwar (approved):
LGTM.
I see we are using ipdk-io/p4runtime but it's same as opensource p4runtime. Please correct me if I am wrong.

We can plan to move to p4lang/p4runtime in next release if that's the case.

ffoulkes:
> > The p4runtime wheel is only one part of the equation.
> > Stratum and its clients use C++ code generated from the contents of the p4runtime submodule.
> > Does the p4runtime submodule reference need to be updated? If not, why not?
> 
> We have already removed customized p4runtime code from stratum (removed LUT manager) and also update p4runtime submodule to not use customized code instead a latest merge from opensource p4lang/P4runtime. This PR was to cleanup some code which was never removed

In other words, the python wheel and the p4runtime reference have been out of sync for some time. This brings them back into sync by removing the custom wheel. That answers my question.

ffoulkes (approved):
LGTM

n-sandeep:
> LGTM. I see we are using ipdk-io/p4runtime but it's same as opensource p4runtime. Please correct me if I am wrong.
> 
> We can plan to move to p4lang/p4runtime in next release if that's the case.

we cannot, we need it because of some more customized changes specific to meters. which will come in next releases

PR #238, opened 2023-08-02
Implement userguide build target

ffoulkes:
- Add 'userguide' CMake target to build the user guide.

- Correct 'maxdepth' on Functions division to 1. We currently display one level in the main window and let the user expand the list by clicking on the entry in the sidebar.


ffoulkes:
Nuts. This can't be integrated until sphinx has been added to the CI workflow environment.

PR #244, opened 2023-08-08
Formalize support for Stratum dependencies

ffoulkes:
- Consolidate the find_package() commands for the packages
  used by Stratum into StratumDependencies.cmake, and include
  the latter from the top-level cmake listfile.

- Replace library names with namespaced library targets.

- Define WITH_OPENSSL variable to indicate whether to link against
  OpenSSL.

- Display Abseil and gRPC package versions.

- Rework logic to find the paths to the host protobuf compiler
  and the gRPC C++ plugin.

- Extract functions to compile Protobuf files and move them to
  ProtobufCompile.cmake.

### DIFF
--- a/stratum/proto/CMakeLists.txt
+++ b/stratum/proto/CMakeLists.txt
@@ -4,38 +4,6 @@
 # SPDX-License-Identifier: Apache 2.0
 #
 
-include(FindProtobuf)
-
-option(INSTALL_PROTO_FILES "Install .proto files" OFF)
-
-#############################
-# Find Protobuf executables #
-#############################
-
-# The Protobuf compiler and gRPC plugin are host tools, even if we're
-# cross-compiling. NO_CMAKE_FIND_ROOT_PATH ensures that we do not
-# search the cross-compiled binaries for the Protobuf executables.
-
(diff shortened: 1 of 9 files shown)


ffoulkes:
I don't have access to a system that allows me to verify that the code cross-compiles for ACC. I'm working on solving that problem, but it might be quicker if someone else can test it for me.

I have no intention of merging this PR until we've verified that it does not break the ACC build.

stratum/proto/CMakeLists.txt
```diff
@@ -53,112 +22,7 @@ string(JOIN ":" PROTO_IMPORT_PATH
...
-            install(
-                FILES
-                    ${SRC_DIR}/${_file}
-                DESTINATION
-                    ${DOT_PROTO_INSTALL_DIR}/${_path}
-            )
-        endif()
-    endforeach()
-endfunction(generate_grpc_files)
+include(ProtobufCompile)
```
ffoulkes, line 25: This is a simple refactoring. I've moved the definitions of the `generate_proto_files()` and `generate_grpc_files() functions to a separate file (`ProtobufCompile.cmake`), reducing the size of `protobuf/CMakeLists.txt` by ~100 lines

PR #246, opened 2023-08-09
Update ovs submodule reference

ffoulkes:
Synchronize with ipdk-latest


ffoulkes:
Wouldn't you know it? The CODEOWNERS file breaks the OVS build!
```text
The following files are in git but not the distribution:
.github/CODEOWNERSThe following files are in git but not the distribution:
.github/CODEOWNERS
```
I think I need to update a list of files to exclude.

ffoulkes:
> Wouldn't you know it? The CODEOWNERS file breaks the OVS build!
> 
> ```
> The following files are in git but not the distribution:
> .github/CODEOWNERS
> ```
> 
> I think I need to update a list of files to exclude.

Problem solved, as evidenced by **All checks have passed**.

nupuruttarwar (approved):
LGTM

PR #248, opened 2023-08-10
Address issues in os_ver_details.sh

ffoulkes:
- Export OS and VER from `get_os_ver_details` in all cases. (It currently does so in only one case.)
- Export NUM_CORES and NUM_THREADS from `get_num_cores`.


nupuruttarwar (approved):
LGTM

nupurjai (approved):

PR #249, opened 2023-08-11
Set recommended compiler flags

ffoulkes:
- Set C and C++ compiler flags according to the Intel Secure Coding Standards.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -84,9 +84,9 @@ endif()
 # Global compiler settings #
 ############################
 
-add_basic_settings()
-#add_recent_security_settings(${CMAKE_BUILD_TYPE})
-add_legacy_security_settings()
+set_basic_compiler_options()
+set_legacy_security_options()
+#set_extended_security_options()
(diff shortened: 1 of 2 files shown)


ffoulkes:
Unfortunately, throwing in all the compiler flags IPAS recommends makes us incompatible with libraries that aren't built with (some of) the flags we're now specifying. It's going to take a while to figure out which options are causing the problems and find a way to deal with it.

I've worked around it for now by adding support for "legacy" legacy -- replicating the settings we used when cross-compiling for ACC.

In the meantime, this module is a significant improvement over the existing code.

ffoulkes:
I don't have access to a system that allows me to verify that the code cross-compiles for ACC. I'm working on solving that problem, but it might be quicker if someone else can test it for me.

I have no intention of merging this PR until we've verified that it does not break the ACC build.

5abeel (approved):
LGTM

nupuruttarwar (approved):
LGTM

PR #250, opened 2023-08-11
Daemonize infrap4d without krnlmon

ffoulkes:
- In the original implementation, infrap4d without krnlmon consisted of nothing but Stratum. There was no support for running as a daemon.

  This CL implements a lightweight version of infrap4d_main.cc (infrap4d_lite.cc) that includes support for daemonization.

### DIFF
--- a/infrap4d/infrap4d_lite.cc
+++ b/infrap4d/infrap4d_lite.cc
@@ -1,0 +1,32 @@
+// infrap4d_main without krnlmon
+
+// Copyright 2022-2023 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+
+#include "gflags/gflags.h"
+#include "stratum/glue/status/status.h"
+#include "stratum/hal/bin/tdi/main.h"
+
+extern "C"  {
+#include "daemon/daemon.h"
+}
+
+DEFINE_bool(detach, true, "Run infrap4d in detached mode");
+
+int main(int argc, char* argv[]) {
+  // Parse infrap4d command line
+  stratum::hal::tdi::ParseCommandLine(argc, argv, true);
+
+  if (FLAGS_detach) {
+      daemonize_start(false);
+      daemonize_complete();
+  }
+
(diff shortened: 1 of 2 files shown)


ffoulkes:
Ready for testing by @nupurjai

nupurjai (approved):

nupurjai:
Please let me know when this is in. Thanks for taking care of this.

ffoulkes:
> Please let me know when this is in. Thanks for taking care of this.

I'm waiting for approval from another code owner (Nupur U, Sandeep, Suresh). Feel free to light a fire under any of them.

infrap4d/infrap4d_lite.cc
```diff
@@ -0,0 +1,32 @@
+// infrap4d_main without krnlmon
```
vsureshkumarp, line 1: File _infrap4d/infrap4d_main.cc_ also can cover the same functionality if we can manage the _krnlmon_ specific code under feature enablement.

If we follow the newer approach, it might become a maintenance issue latter stages.

Definitely there is no logic or functionality issue here. We need to look at the maintainability of the code.

ffoulkes: That was my original intention, until I discovered just how much code I was going to have to conditionalize.

Heavily conditionalized files are hard to maintain, too. It isn't worth it for this little code.
- _infrap4d_main.cc_ is 50 lines.
- _infrap4d_line.cc_ is 32 lines.

We can always combine them later, if the need arises.

-----

On the design front, there is a great deal to be said for providing a separate _main_ file for each infrap4d target configuration. It makes it easy to add and maintain new configurations (e.g., for SONiC), and to do things like defining target-specific command-line flags (or flags with different default values).

In addition to which, not having to deal with #ifdefs will make life a whole lot more pleasant when we move to Bazel.

We'll have a better idea of how to evolve when we have more targets.

nupuruttarwar (approved):
LGTM

satish153 (approved):
LGTM

PR #253, opened 2023-08-12
Convert P4 driver files to find-modules

ffoulkes:
This CL defines the P4 Driver (SDE) libraries for each target as a
package that can be loaded via the find_package() command.
It also defines an abstract interface to apply the linkage properties
of the SDE to the application that uses it (infrap4d).

- Rename the xxxx-driver.cmake files to FindXxxxDriver.cmake.
  This is part of what makes each of them a find-module

- Replace include() commands with find_package() commands.

- Overhaul file contents.
  - Separate the code that finds the libraries from the code that
    defines the targets.
  - Verify that the find_library() calls succeed. The REQUIRED
    parameter is not available in CMake 3.16.
  - Set the INCLUDE_DIRECTORIES and INTERFACE_LANGUAGE
    properties on the library targets.
  - Mark cache variables as advanced.
  - Implement SDE_VERSION variable.
  - Define libraries in the sde:: namespace.

- Implement an add_xxx_target_libraries() function for each driver,
  to set the link properties a target needs to link with the SDE.
  This abstracts and localizes target-specific SDE details, and it
  simplifies the infrap4d build.

Depends on [krnlmon PR #49](https://github.com/ipdk-io/krnlmon/pull/49).

Note that these modules currently depend on SDE_INSTALL_DIR being
defined. Additional work is required to make them proper find-modules
that can be used independently of this project.


ffoulkes:
If you diff two of the find modules, you'll find that they have a great deal in common. I'd suggest reviewing one file to see if you understand it, and if you have any concerns about what it's then doing, and then either skim the other two files or (better still) diff each of them against the file you originally reviewed.

PR #254, opened 2023-08-12
Refine config-cross-recipe script

ffoulkes:
- Break command-line help into Paths and Options sections.

- Adopt recent conventions from make-all.sh.

- Change ES2K_TARGET=YES to TDI_TARGET=ES2K.

- Remove unnecessary echo statements.

- Address shellcheck issues

- Replace old script in top-level directory with symlink to
  revised script in scripts/es2k.


vsureshkumarp (approved):
Changes looks good to me.

config-cross-recipe.sh
```diff
@@ -1,156 +0,0 @@
-#!/bin/bash
```
n-sandeep, line 1: now ACC scripts are modified to pick config-cross-recipe.sh from scripts/es2k/config-cross-recipe.sh
Older reference can be removed

ffoulkes: Cool! Can you approve the PR, so I can merge it?
I'll remove the symlinks in a follow-up commit.

nupuruttarwar (approved):
LGTM

PR #256, opened 2023-08-12
Refine make-cross-ovs script

ffoulkes:
- Adopt recent conventions from make-all.sh.

- Address shellcheck issues.

- Replace old script in top-level directory with symlink to the
  revised script in scripts/es2k.


vsureshkumarp (approved):
Changes looks good to me.

PR #258, opened 2023-08-18
Restore support for Tofino infrap4d build

ffoulkes:
- tofino_main does not support the infrap4d interface, so it won't build with infrap4d_lite. There's no call for us to implement the interface at this time, and I don't want to add conditionals to infrap4d_lite. Just revert to the old way of building the Tofino target.


nupuruttarwar (approved):
LGTM

n-sandeep (approved):
LGTM

PR #260, opened 2023-08-20
Improve make-all.sh helper script

ffoulkes:
- Implement --config option to configure without building.

- Implement --cxx option to specify C++ standard.

- Swap short and long option names in case statement to make it easier to alphabetize by long name.

- Address shellcheck issues.


make-all.sh
```diff
@@ -296,6 +291,6 @@ fi
 
 # Now build the rest of the recipe.
 config_recipe
-if [ ${_CFG_ONLY} -ne 0 ]; then
+if [ ${_CFG_ONLY} -eq 0 ]; then
```
nupuruttarwar, line 294: not equals 0?

ffoulkes: `_CFG_ONLY` is 0 if we want to configure and build, and non-zero if we want to configure without building.
I got the sense wrong when I originally coded it.

nupuruttarwar (approved):

5abeel (approved):
LGTM

PR #266, opened 2023-08-29
Update setup guides to include DEPEND_INSTALL env var

5abeel:
Setup guides had details setting DEPEND_INSTALL environment variable set after usage in previous section. Updating to re-order to set the DEPEND_INSTALL environment variable, and using it in the cmake command

### DIFF
--- a/docs/guides/setup/dpdk-setup-guide.md
+++ b/docs/guides/setup/dpdk-setup-guide.md
@@ -29,8 +29,10 @@ pip3 install -r requirements.txt
 git clone --recursive https://github.com/ipdk-io/networking-recipe.git ipdk.recipe
 cd ipdk.recipe
 export IPDK_RECIPE=`pwd`
+export DEPEND_INSTALL=<absolute path for installing dependencies>
+export SDE_INSTALL=<absolute path for p4 sde install built in previous step>
 cd $IPDK_RECIPE/setup
-cmake -B build -DCMAKE_INSTALL_PREFIX=<dependency install path> [-DUSE_SUDO=ON]
+cmake -B build -DCMAKE_INSTALL_PREFIX=$DEPEND_INSTALL [-DUSE_SUDO=ON]
 cmake --build build [-j<njobs>]
 ```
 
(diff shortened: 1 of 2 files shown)


ffoulkes (commented):
I have some reservations about doing this.

In the nomenclature used by the cmake listfiles, INSTALL is used for _input_ install trees. PREFIX is used for _output_ install trees.

Things get hairier when you're building for ES2K. There, you have _two_ sets of dependency libraries: the _host_ dependencies, which are built for the development system (nominally x86), and the _target_ dependencies, which are built for the runtime system (nominally aarch64).

When we're building in the native environment, we use the same dependency libraries for both development and runtime systems. This is also true for ES2K if we are running P4 Control Plane on one of the x86 host cores. In both cases, DEPEND_INSTALL provides the path to the x86 dependencies.

When we're cross-compiling for the ACC:

- The environment variable for the host dependencies is `HOST_INSTALL`. It is the path to the x86 libraries.
- The environment variable for the target dependencies is `DEPEND_INSTALL`. It is the path to the aarch64 libraries.

You have to build the dependencies twice for cross-compilation. We provide a pair of helper scripts for the ES2K dependency builds, `make-host-deps.sh` and `make-cross-deps.sh`. You have to build the host dependencies first, because you need them to build the target dependencies.

The chances for confusion are high. To try to reduce this confusion, the helper scripts use distinct terminology:

- `--deps` (`-D`) for the input target dependency tree, or the sole dependency tree if you're building in native mode
- `--host` (`-H`) for the input host dependencies, used only when cross-compiling
- `--prefix` (`-P`) for the output install tree

I make a point of using actual file paths (or, in some cases, purpose-specific environment variables) for `CMAKE_INSTALL_PREFIX` in the examples I create. You can damage things if you write to the wrong location. Better to be explicit.

I'm worried that using DEPEND_INSTALL as an output variable will muddy the waters.

ffoulkes:
```text
dfoster@dgfoster-mobl1:~/latest$ ./make-all.sh -h

Configure and build P4 Control Plane software

Paths:
  --deps=DIR       -D  Target dependencies directory [/opt/deps]
  --host=DIR       -H  Host dependencies directory []
  --ovs=DIR        -O  OVS install directory [ovs/install]
  --prefix=DIR     -P  Install directory prefix [install]
  --sde=DIR        -S  SDE install directory [/home/dfoster/mev-p4-sde/install]
  --staging=DIR        Staging directory prefix []
  --toolchain=FILE -T  CMake toolchain file

[redacted]

Environment variables:
  CMAKE_TOOLCHAIN_FILE - Default toolchain file
  DEPEND_INSTALL - Default target dependencies directory
  HOST_INSTALL - Default host dependencies directory
  OVS_INSTALL - Default OVS install directory
  SDE_INSTALL - Default SDE install directory
```

PR #269, opened 2023-09-03
Document the make-all.sh helper script

ffoulkes:


n-sandeep (approved):
LGTM

nupuruttarwar (approved):
LGTM

docs/scripts/make-all.rst
```diff
@@ -0,0 +1,181 @@
...
+
+``--minsize``  
+  Build with ``-DCMAKE_BUILD_TYPE=MinSizeRel``.
+  The compiler settings will default to ``-Os -DNDEBUG``.
+
+``--reldeb``
+  Build with ``-DCMAKE_BUILD_TYPE=RelWithDebInfo``.
+  The compiler settings will default to ``-O2 -g -DNDEBUG``.
+
+``--release``
```
nupuruttarwar, line 146: What will be the default value if nothing is specified? I am trying to understand if we need to modify script where CI images are built, or default config is the one that is without symbols and has less memory footprint?

ffoulkes: If nothing is specified, the helper scripts should not specify CMAKE_BUILD_TYPE when they configure the build, letting the cmake listfile use its own default value.

The default value has been RelWithDebInfo for a long time.

I've been thinking lately that we ought to change the default to Release, which will boost the optimization level from `-O2` to `-O3` in addition to omitting the symbols. In a production system, the default build type _ought_ to be Release. Among other things, the Intel security guide dictates that you use some additional compiler settings to make the resulting images less vulnerable.

(We're not quite there yet -- it turns out that one or more of the recommended settings cause in link-time errors if the options differ between our code and the libraries we link to. It's going to take trial and error to figure out which of the options are causing the problem and what we need to do differently when we build the libraries -- assuming that we're the ones who build them.)

Making this change would affect anybody who expect to be able to run `gdb`. They'd have to start specifying `--reldeb` in order to maintain parity. We would therefore want to (a) announce the planned change in advance, and then (b) advertise the change prominently when we make it.

Another way to reduce the file size is to run `strip` on the image after building it with `RelWithDebInfo`. There is also a way to have the toolchain write the debug symbols to a separate file, which you can load or reference in your debugger. Neither of these would give you `-O3`.

ffoulkes: Good point, though. The document should say what the default behavior is.

PR #275, opened 2023-09-09
Edit make-all.sh user guide

ffoulkes:
- Document what the default build type is if no configuration is specified.

- Replace &reg; with the UTF-8 symbol. RST doesn't render html character entities. (sigh)


docs/scripts/make-all.rst
```diff
@@ -147,6 +147,10 @@ Configurations
   Build with ``-DCMAKE_BUILD_TYPE=Release``.
   The compiler settings will default to ``-O3 -DNDEBUG``.
 
+If no configuration is specified, the CMake listfile will default to
+``RelWithDebInfo``.
+This will probably be changed to ``Release`` in a future release.
```
5abeel, line 152: Dont think we need this that something "will probably be changed to". 

This can be updated when we change the deault

ffoulkes: Actually, I think we should make a _point_ of telling people in advance that we expect something to change.

Especially because we do such a poor job of telling people _after_ we've made the change.

nupuruttarwar (approved):
LGTM

PR #280, opened 2023-09-13
Minor documentation edits

ffoulkes:
- Correct the config-cross-recipe flow diagram to show input from the ovs install tree, not the ovs build tree.

- Change "ES2K Build" to "ACC Build" in the index page for the helper scripts.


docs/scripts/helper-scripts.rst
```diff
@@ -13,8 +13,8 @@ Build
 
    make-all
 
-ES2K Build
-----------
+ACC Build
```
nupuruttarwar, line 16: Suggestion: Since the helper-scripts.rst is part of general documentation, it will be good to have something that indicates that ACC build is applicable only for ES2K target.

Maybe change it to ES2K ACC build?

ffoulkes: I'll do you one better. Instead of changing the heading, I'll add an explanatory sentence that says what ACC means. :)

nupuruttarwar (approved):
LGTM

5abeel (approved):
LGTM

PR #281, opened 2023-09-13
Clarify nature of es2k setup guide

ffoulkes:
- Say at the top of the document that it is for a host build.

- Rewrite the "Install P4 SDE" section to account for its vagueness and disclaim responsiblity for documenting the setup procedures for the ES2K.


ffoulkes:
This is a step in the right direction. It is not the final solution. But it provides the reader with answers to two of the most pressing questions, and it does so now.

5abeel (approved):
LGTM

n-sandeep (approved):
LGTM

nupuruttarwar (approved):
LGTM

PR #286, opened 2023-09-19
Let cmake supply default build type

ffoulkes:
- In CMakeLists.txt, define default CMAKE_BUILD_TYPE after project() command. Provide list of valid build types for user interface tools.

- In make-all.sh, don't define CMAKE_BUILD_TYPE if type is not explicitly specified.

- In make-cross-ovs.sh and config-cross-recipe.sh, don't define default CMAKE_BUILD_TYPE.

- Separate Path options in the config-cross-recipe.sh help text and user guide into Host and Target categories.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,9 +7,6 @@
 # Version 3.15 is the baseline for P4 Control Plane.
 cmake_minimum_required(VERSION 3.15)
 
-# CMAKE_BUILD_TYPE must be set before the first project() command.
-set(CMAKE_BUILD_TYPE "RelWithDebInfo" CACHE STRING "Type of build to perform")
-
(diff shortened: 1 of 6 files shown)


CMakeLists.txt
```diff
@@ -7,9 +7,6 @@
 # Version 3.15 is the baseline for P4 Control Plane.
 cmake_minimum_required(VERSION 3.15)
 
-# CMAKE_BUILD_TYPE must be set before the first project() command.
```
5abeel, line 10: Is this forcing users to specify the build type during compilation (without picking a default?)
A `make-all.sh` should run with some default configuration

5abeel: Ack

5abeel (approved):
LGTM

nupuruttarwar (approved):
LGTM

PR #301, opened 2023-09-29
Fix cmake configuration error

ffoulkes:
- As the result of the most recent changes, the cmake configuration step fails because the gflags::gflags_shared target is undefined. Address this by finding the gflags package in the top-level cmake listfile.


5abeel (approved):
LGTM

nupuruttarwar (approved):
LGTM

PR #303, opened 2023-09-30
Formalize logic to import Stratum dependencies

ffoulkes:
- Consolidate the find_package() commands for the packages used by Stratum into StratumDependencies.cmake, and include the latter from the top-level listfile.

- Replace library names with namespaced library targets.

- Define WITH_OPENSSL variable to specify whether to link against OpenSSL.

- Display Abseil and gRPC package versions

- Rework logic to find the paths to the host protobuf compiler and the gRPC C++ plugin.

- Set the ADVANCED attribute on cache variables that should not normally be displayed in cmake GUIs.

Supersedes https://github.com/ipdk-io/networking-recipe/pull/244.


5abeel (approved):
LGTM

satish153 (approved):
LGTM

PR #305, opened 2023-09-30
Overhaul cmake interface to DPDK SDE

ffoulkes:
- Replace the dpdk-driver.cmake include file with FindDpdkDriver.cmake, a find-module that allows us to use `find_package(DpdkDriver)` to define namespaced cmake targets for the DPDK SDE.

- Define an `add_dpdk_target_libraries()` function that provides an abstract interface to set the properties a target needs to compile and link with the DPDK SDE.

Depends on https://github.com/ipdk-io/krnlmon/pull/49.


5abeel (approved):
LGTM

PR #306, opened 2023-09-30
Overhaul cmake interface to ES2K SDE

ffoulkes:
- Replace the es2k-driver.cmake include file with FindEs2kDriver.cmake, a find-module that allows us to use `find_package(Es2kDriver)` to define namespaced cmake targets for the ES2K SDE.

- Define an `add_es2k_target_libraries()` function that provides an abstract interface to set the properties a target needs in order to compile and link with the ES2K SDE.

Depends on https://github.com/ipdk-io/krnlmon/pull/49.


ffoulkes:
_Nota bene:_ The CI build failure is a consequence of the original PR having been separated into three commits.

CI does a *DPDK* build on an *ES2K* commit. The build fails because the (unmodified) DPDK driver package does not define the target names the (modified) version of krnlmon expects.

cmake/FindEs2kDriver.cmake
```diff
@@ -0,0 +1,151 @@
+# FindEs2kDriver.cmake - import ES2K P4 Driver (SDE).
+#
+# Copyright 2022-2023 Intel Corporation
```
5abeel, line 3: New file shouldnt have the 2022 copyright(?)

cmake/FindEs2kDriver.cmake
```diff
@@ -0,0 +1,151 @@
...
+
+#-----------------------------------------------------------------------
+# Define SDE_INCLUDE_DIR
+#-----------------------------------------------------------------------
+find_path(SDE_INCLUDE_DIR
+    NAMES "bf_switchd/bf_switchd.h"
+    PATHS ${SDE_INSTALL_DIR}/include
+)
+if(NOT SDE_INCLUDE_DIR)
+  message(FATAL_ERROR "bf_switchd.h not found")
```
5abeel, line 18: Why are we looking specifically for the bf_switchd.h? Or is it just one file that was picked as that was common across all targets?

cmake/FindEs2kDriver.cmake
ffoulkes: I suppose I could consider this to be an original file. It's a complete rewrite of `es2k-driver.cmake`, with very few similarities (and those so obvious that one could hardly claim to have created them).

cmake/FindEs2kDriver.cmake
ffoulkes: Just looking for something to verify that SDE_INSTALL_DIR points to the root of an SDE install tree and that seemed unlikely to change.

5abeel (approved):
LGTM

PR #307, opened 2023-09-30
Overhaul cmake interface to Tofino SDE

ffoulkes:
- Replace the tofino-driver.cmake include file with FindTofinoDriver.cmake, a find-module that allows us to use `find_package(TofinoDriver)` to define namespaced cmake targets for the Tofino SDE.

- Define an `add_tofino_target_libraries()` function that provides an abstract interface to set the properties a target needs in order to compile and link with the Tofino SDE.


nupuruttarwar (approved):
LGTM

5abeel (approved):
LGTM

PR #311, opened 2023-10-09
Update find_program() commands

ffoulkes:
- Specify NO_CMAKE_FIND_ROOT_PATH when searching for the Protobuf compiler and gRPC plugin. We do this to ensure that we search the host depencies, not the target dependencies, when cross-compiling.

If https://github.com/ipdk-io/networking-recipe/pull/306 causes problems with cross-compilation, this PR should fix them.


5abeel (approved):
LGTM

nupuruttarwar (approved):
LGTM

PR #310, opened 2023-10-06
Have es2k package add sde to search path

ffoulkes:
- Modify FindEs2kDriver to add SDE_INSTALL_DIR to CMAKE_PREFIX_PATH or CMAKE_FIND_ROOT_PATH.

PR #315, opened 2023-10-14
 Update driver find-modules to update search path

ffoulkes:
- Modify DPDK and Tofino Driver modules to add SDE_INSTALL_DIR to CMAKE_PREFIX_PATH.

This CL does for DPDK and Tofino what https://github.com/ipdk-io/networking-recipe/pull/310 did for ES2K. It enables standalone builds of the Kernel Monitor for DPDK, and will allow us to perform include-what-you-use analysis (https://github.com/ipdk-io/krnlmon/pull/61).


5abeel (approved):
LGTM

PR #319, opened 2023-10-20
Append SDE pkgconfig folders to PKG_CONFIG_PATH

ffoulkes:
- Modify FindDpdkDriver module to check for `pkgconfig` directories in the SDE and append them to PKG_CONFIG_PATH before using pkg-config to find the DPDK module.

This should eliminate the need to set PKG_CONFIG_PATH in the environment and allow the build to "just work".

### DIFF
--- a/cmake/FindDpdkDriver.cmake
+++ b/cmake/FindDpdkDriver.cmake
@@ -8,12 +8,6 @@
 # It needs additional work to make it a proper CMake find-module.
 
 #-----------------------------------------------------------------------
-# Use pkg-config to find DPDK module
-#-----------------------------------------------------------------------
-include(FindPkgConfig)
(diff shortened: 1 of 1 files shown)


nupuruttarwar (approved):
Great find! Change LGTM for DPDK. Can we extend it to ES2K and Tofino as well?
Also, can we append the path for dependencies as well?

ffoulkes:
> Great find! Change LGTM for DPDK. Can we extend it to ES2K and Tofino as well? Also, can we append the path for dependencies as well?

To the best of my knowledge, this change only applies to finding pkg-config files that are part of the SDE (i.e., DPDK for the DPDK TDI target).

5abeel (approved):
LGTM

PR #320, opened 2023-10-24
Replace networking-recipe CI pipeline

ffoulkes:
- Replace the P4 Control Plane CI pipeline. The new pipeline uses precompiled copies of the DPDK SDE and stratum dependencies instead of building them from scratch every time, and it moves the job to run the Stratum unit tests to the stratum repository.

  The net effect is to reduce the run time of the networking-recipe commit check from ~1.2-1.5 hours to ~10-15 minutes.


nupuruttarwar (approved):
Great improvement! LGTM

.github/workflows/pipeline.yml
```diff
@@ -0,0 +1,150 @@
...
+  group: build-${{ github.ref }}
+  cancel-in-progress: true
+
+permissions: read-all
+
+env:
+  #---------------------------------------------------------------------
+  # Environment variables
+  #---------------------------------------------------------------------
+  SDE_REPOSITORY: ipdk-io/p4dev.dpdk-sde
```
5abeel, line 24: Is this a sandbox name? We should probably not have a name with `dev` in it?

.github/workflows/pipeline.yml
```diff
@@ -0,0 +1,150 @@
...
+  SDE_TAG: 2023.10.1
```
5abeel, line 25: Where does this tag come from?

ffoulkes: The answer to your question may be found here:

- [dpdk-sde 2023.10.1](https://github.com/ipdk-io/p4dev.dpdk-sde/releases/tag/2023.10.1)

I've scheduled a CI Pipeline Review (Teams meeting) for 9:00 am Wednesday to discuss the mechanics of the revised pipeline. I figured a guided tour would be easier than a written explanation.

5abeel (approved):
LGTM

PR #322, opened 2023-10-29
P4CP configuration file generator

ffoulkes:
Command-line utility to generate cmake configuration files for building P4 Control Plane.

This is a convenient alternative to environment variables and helper scripts. which become rather tedious when you find yourself typing the same command-line flags every you run (and even more so when you find yourself switching between the same two or three configurations).

```bash
cmake -B build -C es2k-setup.cmake
cmake --build build -j6 --target install
```

The utility can also save default parameters to a file that it loads on subsequent runs, and it can generate files to define common environment variables.

It does not yet support cross-compilation.


ffoulkes:
I'm think of dropping the `.py` suffix and making the command just `p4cpconfig`. Any opinions?

PR #325, opened 2023-11-02
Using copy_config_files.sh under es2k

satish153:
**Three changes in this PR:**
1. `copy_config_files.sh.in` under scripts/common has content that is target-specific. It needs cleaning up. 
For now, `copy_config_files.sh` under  scripts/es2k is to be used for copying files for es2k target. 
2. A minor fix in copy_config_files.sh.in
3. Installing tdi_pktio.json from SDE_INSTALL path. This file is needed for packetIO feature.

### DIFF
--- a/scripts/common/copy_config_files.sh.in
+++ b/scripts/common/copy_config_files.sh.in
@@ -41,6 +41,6 @@ sudo cp ${SOURCE_DIR}/ca.conf ${TLS_CERTS_DIR}/
 sudo cp ${SOURCE_DIR}/grpc-client.conf ${TLS_CERTS_DIR}/
 sudo cp ${P4CP_INSTALL}/sbin/generate-certs.sh ${TLS_CERTS_DIR}/
 
-sudo mkdir /usr/share/target_sys/
+sudo mkdir -p /usr/share/target_sys/
 sudo cp $SDE_INSTALL/share/target_sys/zlog-cfg /usr/share/target_sys/
 set +e
--- a/scripts/es2k/CMakeLists.txt
+++ b/scripts/es2k/CMakeLists.txt
@@ -4,18 +4,9 @@
 # SPDX-License-Identifier: Apache 2.0
 #
 
-# Customize copy_config_files.sh script
-
-set(TARGET_NAME "es2k")
-
-configure_file("../common/copy_config_files.sh.in"
-               ${CMAKE_CURRENT_BINARY_DIR}/copy_config_files.sh @ONLY)
-
-unset(TARGET_NAME)
-
 install(
     PROGRAMS
-        ${CMAKE_CURRENT_BINARY_DIR}/copy_config_files.sh
+        copy_config_files.sh
     TYPE
         SBIN
 )
--- a/scripts/es2k/copy_config_files.sh
+++ b/scripts/es2k/copy_config_files.sh
@@ -18,17 +18,32 @@ then
     return 0
 fi
 
-export IPDK_INSTALL=$1
+export P4CP_INSTALL=$1
 export SDE_INSTALL=$2
 
+SOURCE_DIR=${P4CP_INSTALL}/share/stratum/es2k
+TARGET_DIR=/usr/share/stratum/es2k
+TLS_CERTS_DIR=/usr/share/stratum
+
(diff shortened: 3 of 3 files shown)


n-sandeep:
Currently the script "copy_config_files.sh" that copies under $P4CP_INSTALL is from "scripts/**common**/copy_config_files.sh.in", we are not using "scripts/**es2k**/copy_config_files.sh" i believe?

satish153:
> Currently the script "copy_config_files.sh" that copies under $P4CP_INSTALL is from "scripts/**common**/copy_config_files.sh.in", we are not using "scripts/**es2k**/copy_config_files.sh" anywhere I believe.
> 
> configure_file("../common/copy_config_files.sh.in" ${CMAKE_CURRENT_BINARY_DIR}/copy_config_files.sh @only)

Yes, that script is a template from which target-specific script is generated. However since it is a common file, it is not correct to have target-specific commands in there. 
Eg: I want to copy tdi_pktio.json only for es2k. 
I could accomplish it by adding 'if' checks. But there is already a target-specific file under the es2k directory, so I used it and made changes in CMakeFiles.txt to copy it to the install directory.

n-sandeep:
> > Currently the script "copy_config_files.sh" that copies under $P4CP_INSTALL is from "scripts/**common**/copy_config_files.sh.in", we are not using "scripts/**es2k**/copy_config_files.sh" anywhere I believe.
> > configure_file("../common/copy_config_files.sh.in" ${CMAKE_CURRENT_BINARY_DIR}/copy_config_files.sh @only)
> 
> Yes, that script is a template from which target-specific script is generated. However since it is a common file, it is not correct to have target-specific commands in there. Eg: I want to copy tdi_pktio.json only for es2k. I could accomplish it by adding 'if' checks. But there is already a target-specific file under the es2k directory, so I used it and made changes in CMakeFiles.txt to copy it to the install directory.

So with your changes we are going to have "scripts/**es2k**/copy_config_files.sh" installed in $P4CP_INSTALL. Then we need to add TLS related configs present in "scripts/**common**/copy_config_files.sh.in" as well in  "scripts/**es2k**/copy_config_files.sh" file.

ffoulkes:
The purpose of this PR is **not** to "use copy_config_files.sh under es2k".

Its purpose is to modify copy_config_files.sh to install tdi_pktio.json on es2k.

The reengineering of copy_config_files.sh is a crude means to that end.

satish153:
> > > Currently the script "copy_config_files.sh" that copies under $P4CP_INSTALL is from "scripts/**common**/copy_config_files.sh.in", we are not using "scripts/**es2k**/copy_config_files.sh" anywhere I believe.
> > > configure_file("../common/copy_config_files.sh.in" ${CMAKE_CURRENT_BINARY_DIR}/copy_config_files.sh @only)
> > 
> > 
> > Yes, that script is a template from which target-specific script is generated. However since it is a common file, it is not correct to have target-specific commands in there. Eg: I want to copy tdi_pktio.json only for es2k. I could accomplish it by adding 'if' checks. But there is already a target-specific file under the es2k directory, so I used it and made changes in CMakeFiles.txt to copy it to the install directory.
> 
> So with your changes we are going to have "scripts/**es2k**/copy_config_files.sh" installed in $P4CP_INSTALL. Then we need to add TLS related configs present in "scripts/**common**/copy_config_files.sh.in" as well in "scripts/**es2k**/copy_config_files.sh" file.

Oh yes!! I missed that. Thanks for noticing it. Added them now.

ffoulkes:
May I offer [PR #327](https://github.com/ipdk-io/networking-recipe/pull/327) as an alternative? It uses a template engine instead of simple text substitution, which allows us to create templates with conditional logic.

The `genscript.py` utility requires `pip install Jinja2` on the development system to work.

According to usage text the shell script displays:
```text
    echo " - Usage: sudo copy_config_files.sh <P4CP_INSTALL> <SDE_INSTALL>"
```
the user is expected to run the using `sudo` (or else, presumably, be running as `root`). This makes perfect sense. The whole operation requires root access. It's why we put the script in `sbin` instead of `bin`.

This means there's no need to specify `sudo` on every flipping command in the script.

I have therefore removed them from the template.

PR #327, opened 2023-11-04
Use template to generate copy_config_files.sh

ffoulkes:
- Create a Jinja template and Python driver to generate the copy_config_files.sh file. This gives us more flexibility the cmake configure_files() command.

I'm offering this as an alternative to PR https://github.com/ipdk-io/networking-recipe/pull/325


ffoulkes:
The `genscript.py` utility requires `pip install Jinja2` on the development system to work.

-----

According to usage text the shell script displays:
```text
    echo " - Usage: sudo copy_config_files.sh <P4CP_INSTALL> <SDE_INSTALL>"
```
the user is expected to run the script using `sudo` (or else be running as `root`). This makes perfect sense. The purpose of the script is to update system directories. It inherently requires administrative privilege. It's why we put the script in `sbin` instead of `bin`.

This means there's no need to specify `sudo` on every flipping command in the script.

I have therefore removed the `sudo` commands from the template.

----

I've removed the `set +e` from the end of the script. The `-e` option is reset to its previous value when the script exits. (This is _not_ the case for scripts you `source`.)

satish153 (approved):
This is nice. Tested it for es2k and dpdk targets.

n-sandeep:
> `Jinja2`

So, we need to update documents/requirements file to tell user that he needs to install "Jinja2" before building networking-recipe?
If Yes, this package need to be added in ACC python packages list as well.

ffoulkes:
> > `Jinja2`
> 
> So, we need to update documents/requirements file to tell user that he needs to install "Jinja2" before building networking-recipe? If Yes, this package need to be added in ACC python packages list as well.

That's correct.

Do we have a requirements document or standard requirements file? Where do I find it/them?

n-sandeep:
> > > `Jinja2`
> > 
> > 
> > So, we need to update documents/requirements file to tell user that he needs to install "Jinja2" before building networking-recipe? If Yes, this package need to be added in ACC python packages list as well.
> 
> That's correct.
> 
> Do we have a requirements document or standard requirements file? Where do I find it/them?

We have a requirements.txt file under networking-recipe, but looks like not updated at all. Better to start updating that file for all python pakages I believe.

nupuruttarwar (commented):
Logic LGTM. Please update requirements.txt to include jinja2. Also, is the PR to include jinja2 in list of ACC packages already in review?

ffoulkes:
> Logic LGTM. Please update requirements.txt to include jinja2. Also, is the PR to include jinja2 in list of ACC packages already in review?

I've added the updated requirements.txt file to the PR.

Jinja2 isn't an ACC requirement. It runs on the development host at build time. I believe there's a Python script to install the build dependencies.

nupuruttarwar (approved):
LGTM

requirements.txt
```diff
@@ -1 +1,2 @@
 p4runtime==1.3.0
+Jinja2
```
5abeel, line 2: It is recommended to pin the version down for a repeatable/dependable build outcomes.

Some linters like Docker linters (which this is not), fail when versions are not pinned down

PR #340, opened 2023-11-14
enabled dependabot for pip/github-action packages

ipdksw:
PR to enable depondabot scan for pip and github actions
More info: https://docs.github.com/en/code-security/getting-started/dependabot-quickstart-guide

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -2,7 +2,7 @@
 # package ecosystems to update and where the package manifests are located.
 # Please see the documentation for all configuration options:
 # https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
-
+---
 version: 2
 updates:
   - package-ecosystem: "pip" # See documentation for possible values
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,15 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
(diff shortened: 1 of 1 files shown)


ffoulkes (approved):
The PR should include a changelist comment explaining the purpose / nature of the change.

I've been using YAML for a number of years, but I had to look up the meaning of three hyphens.

They're really not appropriate here. The hyphens mark the start of a document when you need a separator. The comment block is documentation for the YAML that follows, so it's part of the same document.

Harmless in this case, but still inappropriate.

ipdksw:
> The PR should include a changelist comment explaining the purpose / nature of the change.
> 
> I've been using YAML for a number of years, but I had to look up the meaning of three hyphens.
> 
> They're really not appropriate here. The hyphens mark the start of a document when you need a separator. The comment block is documentation for the YAML that follows, so it's part of the same document.
> 
> Harmless in this case, but still inappropriate.

Thanks Derek, updated the PR description!

5abeel (approved):
LGTM

Just one minor comment: Do we need to check for daily updates? pip packages are not updated that frequently.

mgaravin:
> LGTM
> 
> Just one minor comment: Do we need to check for daily updates? pip packages are not updated that frequently.

@5abeel : will change the frequency once we have report for SDLe.

ffoulkes:
Hi Aravind,

You should probably enable Dependabot for github actions on:

  *   ipdk-io/krnlmon
  *   ipdk-io/stratum-deps
  *   idpk-io/stratum-dev

What’s the scope of the pip scanning?

  *   pip commands in github workflows?
  *   Files in the repository (and, if so, what kinds of files)?

Best regards,
Derek

Issue #336, opened 2023-11-12
Which requirements.txt file is the one mentioned in these instructions?  And does it have an error in its contents?


jafingerhut:
On this page: https://ipdk.io/p4cp-userguide/guides/setup/es2k-setup-guide.html

It recommends running the command `pip3 install -r requirements.txt` quite early on the page.  However, it does not say _which_ `requirements.txt` file it should be.

Is it the one in the root directory of the repo https://github.com/ipdk-io/networking-recipe ?  If yes, not that the page linked above only mentions cloning that repo many many steps later.  It would be good to clarify this somehow, perhaps by giving the instructions to clone the networking-recipe repo, and change to its root directory, before running the `pip3 install -r requirements.txt` command.

In addition, that `requirements.txt` file contains lines `fcntl` and `select`.

I get the following error when I run the `pip3 install -r requirements.txt` command on that file, on all of Ubuntu 20.04 and 22.04, and Fedora 37 and 38 systems:

```
$ sudo pip3 install -r requirements.txt 
Collecting p4runtime==1.3.0
  Downloading p4runtime-1.3.0-0-py3-none-any.whl (37 kB)
Collecting Jinja2>=3.0.0
  Downloading Jinja2-3.1.2-py3-none-any.whl (133 kB)
     |████████████████████████████████| 133 kB 3.0 MB/s 
ERROR: Could not find a version that satisfies the requirement fcntl (from -r requirements.txt (line 3)) (from versions: none)
ERROR: No matching distribution found for fcntl (from -r requirements.txt (line 3))
```

and when that error appears, the command installs _no_ Python packages at all, not even the ones that exist.

If I delete the lines `fcntl` and `select`, the `pip3` command does successfully install the other 2 packages.  As far as I can tell, the `fcntl` and `select` packages are installed by default with Python3 on all of those operating systems.

PR #345, opened 2023-11-15
Remove invalid packages from requirements.txt

ffoulkes:
- Remove `fcntl` and `select`, neither of which is available on pypi.org.

  https://github.com/ipdk-io/networking-recipe/issues/336


satish153 (approved):
Thanks for fixing this.
I realized yesterday that these modules are part of the python standard library.

5abeel (approved):
LGTM

PR #347, opened 2023-11-16
Improve GitHub workflow

ffoulkes:
- Do a recursive clone of networking-recipe in krnlmon_unit_tests. We need to fetch both krnlmon and SAI.

- Don't install Go for krnlmon_unit_tests. We only need it when building the p4runtime protobufs.


5abeel (approved):
LGTM

nupuruttarwar (approved):

PR #350, opened 2023-11-16
Support p4runtime protobufs when cross-compiling

ffoulkes:
- Fix parse error in HostProtoC when version string returned by protoc is empty.

- Pass HOST_DEPEND_DIR to external project when building protobufs through networking-recipe.

- Revise protobufs CMakeLists.txt file to use HOST_DEPEND_DIR instead of DEPEND_INSTALL_DIR when cross-compiling.


ffoulkes:
On further reflection, I think this is a bad idea. Building the P4Runtime protobufs is inherently a host-mode operation. We don't use a single cross-compilation tool.

Instead of bending over backwards to try to make this work in the cross-compilation environment (it still has problems), we should simplify things by returning an error. It would make the code simpler. It would make the instructions for using it simpler.

You use host mode to build the protobufs, just as you use host mode to build the host dependencies.

PR #351, opened 2023-11-17
Refine p4runtime protobuf generation

ffoulkes:
- Fix parse error in HostProtoC when version string returned by protoc is empty.

- It turns out that there is no effective difference between building as an external project (through networking-recipe) and build when invoked directly in the protobufs directory. Revise project initialization to take this into account.

- Protobuf generation does not use any cross-compilation tools. Don't support it when cross-compiling.

### DIFF
--- a/cmake/FindHostProtoc.cmake
+++ b/cmake/FindHostProtoc.cmake
@@ -13,19 +13,31 @@ include_guard(GLOBAL)
 find_program(HOST_PROTOC_COMMAND "protoc" NO_CMAKE_FIND_ROOT_PATH)
 mark_as_advanced(HOST_PROTOC_COMMAND)
 
-if(HOST_PROTOC_COMMAND)
-  execute_process(
-    COMMAND ${HOST_PROTOC_COMMAND} --version
-    OUTPUT_VARIABLE protoc_output
-  )
-  string(STRIP "${protoc_output}" protoc_output)
-  string(REGEX REPLACE
-    "^libprotoc +([0-9.]+)$" "\\1" PROTOC_VERSION ${protoc_output})
-  message(STATUS "Found protoc: ${HOST_PROTOC_COMMAND} (version found \"${PROTOC_VERSION}\")")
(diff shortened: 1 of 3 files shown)


bharticemk (approved):
LGTM

satish153 (approved):

cmake/FindHostProtoc.cmake
```diff
@@ -13,20 +13,32 @@ include_guard(GLOBAL)
...
-    COMMAND ${HOST_PROTOC_COMMAND} --version
-    OUTPUT_VARIABLE protoc_output
-  )
-  string(STRIP "${protoc_output}" protoc_output)
-  string(REGEX REPLACE
-    "^libprotoc +([0-9.]+)$" "\\1" PROTOC_VERSION ${protoc_output})
-  message(STATUS "Found protoc: ${HOST_PROTOC_COMMAND} (version found \"${PROTOC_VERSION}\")")
-  unset(protoc_output)
-else()
+if(NOT HOST_PROTOC_COMMAND)
```
5abeel, line 16: To clarify, this is only a fatal error when building against protobufs target? Any other cmake error for regular builds should not result in a fatal error.

ffoulkes: It's a fatal error for P4 Control Plane as well. We have to use the host Protobuf compiler to compile our .proto files, regardless of whether we're compiling for the host platform or the ACC.

We should be okay. It's been a fatal error for a long time now. I just changed the logic from
```cmake
if (HOST_PROTOC_COMMAND)
  good stuff
else()
  message(FATAL_ERROR)
endif()
```

to

```cmake
if(NOT HOST_PROTOC_COMMAND)
  message(FATAL_ERROR)
endif()

good stuff
```

5abeel: ok. Since networking-recipe also heavily depends on the protobuf compilation, I guess that makes sense.

5abeel (approved):
LGTM

PR #359, opened 2023-11-22
LAG documentation

aashishkuma:
Documentation for Linux Networking with LAG for ES2K.

PR #362, opened 2023-11-29
Add LNW with LAG to user guide

ffoulkes:
This PR against the `update_lag_guide` PR is needed to make the document part of the user guide.

@aashishkuma - You should be able to merge this into your PR. It does not require any approvals


ffoulkes:
> Might be better to just add this commit to PR #359 or alternatively target merge this into main branch

This PR was against P4 #359. It is set up to be merged into the PR. It requires no approvals, so it's really easy for the original author to merge it into their PR (which is "adding it to the commit"). Merging it into the main branch would be unnecessary extra work.

PR #374, opened 2023-12-10
Fix VLAN TAG issue in original packet.

n-sandeep:
With the existing code, for a VLAN TAG port when we receive a TAG packet, we are appedning one more level of TAG and forwarding the packet. This is causing double VLAN tag on the egress packet. This causes traffic to be dropped.

With this code, If a port is already configured with TAG, then we know that it can receive only TAG traffic and that means original pkt will already have a TAG and we dont need to add one more tag to the packet. Remove piece of code which adds TAG to the orig packet.


vsureshkumarp (approved):
Changes looks good to me.

ffoulkes (approved):
LGTM

PR #376, opened 2023-12-12
Security documentation update

5abeel:
Cleanup security related documentation for clarity

Updates here include:
* Remove the non-default certificate location section. This only works for server side, but client still needs the default location -- for any user running server & client on same machine, this is confusing. Also, in limitations we note that only default location is supported, so it is better to remove this section until full non-default location is supported across server and client apps
* Remove the $IPDK_RECIPE section of commands in examples


ffoulkes (approved):
LGTM

nupuruttarwar (approved):
LGTM

PR #389, opened 2024-01-11
Modify LNW documentation for 24.01 release

n-sandeep:


docs/apps/lnw/es2k/es2k-linux-networking.md
```diff
@@ -16,6 +16,68 @@ To enable this feature we have,
 - `p4rt-ctl`: This python CLI includes a p4runtime client. Programs IPU E2100 with runtime rules by communicating with gRPC server.
 - `Kernel stack`: All underlay related configurations are picked by `kernel monitor` thread via netlink events in `infrap4d` and these are programmed in IPU E2100 by calling TDI front end APIs.
 
+### P4 program changes in 24.01
+
+Modify p4 programm with below changes, to enable support for
```
nupuruttarwar, line 21: P4 program?

docs/apps/lnw/es2k/es2k-linux-networking.md
```diff
@@ -16,6 +16,68 @@ To enable this feature we have,
...
+
+- IPv6 traffic with VxLAN support.
+- VxLAN traffic fails when remote overlay VMs MAC is learnt on OvS, before DUTs overlay VMs MAC address.
```
nupuruttarwar, line 24: It's not clear what are we enabling the support for here. The statement describes the problem but not the functionality. 1st one sounds like feature enhancement but  wasn't IPv6 code already available. Are both bug fixes?

ffoulkes (commented):
### Issue 1

The added section is not part of the **Feature Overview**. It belongs at heading level 2, not heading level 3.

### Issue 2

Patches belong in patch files. Putting them in a document makes the document harder to read and the patch harder to apply. The document should describe the problem(s) being fixed and/or the feature(s) being added, and then link to the patch file(s).

### Issue 3 

The introduction does not make sense.

To understand what it actually says, apply the first sentence to each item in the list that follows, as if you were distributing a term in a system of algebraic equations.

- Modify [the] P4 program with [the following] changes, to enable support for _IPv6 traffic with VxLAN support_.
- Modify [the] P4 program with [the following] changes, to enable support for _VxLAN traffic fails when remote overlay VMs MAC is learnt on OvS, before DUTs overlay VMs MAC address_.

These are independent paragraphs, not bullets in a list.

### Issue 4

The purpose of the second bullet is unclear. Are you describing a defect in the P4 program that the patch fixes, a defect in the P4 program that the patch does not fix, or a defect in the P4 program once the patch has been applied?

If the patch adds a feature and fixes a defect, it should clearly state this.

The second bullet also needs to be rewritten. It doesn't make sense.

ffoulkes (changes requested):
See long comment for issues that need to be addressed (I switched back to code view to check something and GitHub kindly changed the radio button to "Comment".)

docs/apps/lnw/es2k/es2k-lnw-overlay-vms.md
```diff
@@ -45,7 +45,7 @@ System under test will have above topology running the networking recipe. Link P
 
 ## Creating the topology
 
-Follow steps mentioned in [Running Infrap4d on Intel E2100](/guides/es2k/running-infrap4d.md) for starting `infrap4d` process and creating protobuf binary for `fxp-net_linux-networking-v2` p4 program.
+Follow steps mentioned in [Running Infrap4d on Intel E2100](/docs/guides/es2k/running-infrap4d.md) for starting `infrap4d` process and creating protobuf binary for `fxp-net_linux-networking-v2` p4 program. Modify `program-name` in `/usr/share/stratum/es2k/es2k_skip_p4.conf` to `fxp-net_linux-networking-v2` before exploring the usecase.
```
ffoulkes, line 48:
```suggestion
Follow steps mentioned in [Running Infrap4d on Intel E2100](/guides/es2k/running-infrap4d.md) for starting `infrap4d` process and creating protobuf binary for `fxp-net_linux-networking-v2` p4 program. Modify `program-name` in `/usr/share/stratum/es2k/es2k_skip_p4.conf` to `fxp-net_linux-networking-v2` before exploring the usecase.
```
Relative links in Sphinx documents treat `/docs` as the implied root. You must omit **/docs** or the link will be broken in the user guide. Links do not operate the same when you're browsing the .md files as they do in the rendered document.

ffoulkes (changes requested):
You have to omit the `/docs` prefix on internal links, or you'll get a broken link in the user guide.

docs/apps/lnw/es2k/es2k-linux-networking.md
n-sandeep: ACK

docs/apps/lnw/es2k/es2k-linux-networking.md
n-sandeep: IPv6 is already available but due to a defect in p4 program, without this p4 programs change, IPv6 with VxLAN traffic will fail.

nupuruttarwar (approved):
LGTM

ffoulkes (approved):
Much better. Thank you!

LGTM

docs/apps/lnw/es2k/es2k-linux-networking.md
ffoulkes, line 1: **Danger, Will Robinson!**

Please do not create branches whose names begin with `ipdk_v`.

This prefix is reserved for _release_ branches.
There are GitHub workflows that are triggered by branch names with this prefix!

PR #407, opened 2024-02-11
Development Update 24.07

ffoulkes:
First draft of the public-facing P4CP "release" notes.

I've covered the Stratum updates and added a section on the P4Runtime Fork. (Old business, but we should probably mention it in every development update. Future editions can be shorter.)

Does anyone have anything to add? We should include any information about P4CP that's in the MEV release notes, since the general public does not have access to them.

### DIFF
--- a/docs/updates/development-update-24.07.md
+++ b/docs/updates/development-update-24.07.md
@@ -10,19 +10,19 @@ the Intel IPU E2100.
 ### Client Flags
 
 - The Stratum `CredentialsManager` class and `gnmi_cli` tool have been aligned
-  with the current upstream version, and IPDK modifications have been
-  overhauled to be more consistent with Stratum as a whole.
+  with the current upstream versions, and IPDK modifications have been
+  revised to be more consistent with Stratum as a whole.
 
   The `-grpc_client_cert_req_type` flag is now `-client_cert_req_type`, and
-  its value names have been shortened.
+  its values have been shortened. They are also now case-insensitive.
 
   | Old Name | New Name |
   | -------- | -------- |
-  | NO_REQUEST_CLIENT_CERT | NO_REQUEST |
-  | REQUEST_CLIENT_CERT_NO_VERIFY | REQUEST_NO_VERIFY |
-  | REQUEST_CLIENT_CERT_AND_VERIFY | REQUEST_AND_VERIFY |
-  | REQUIRE_CLIENT_CERT_NO_VERIFY | REQUIRE_NO_VERIFY |
-  | REQUIRE_CLIENT_CERT_AND_VERIFY | REQUIRE_AND_VERIFY |
+  | NO_REQUEST_CLIENT_CERT | no_request |
+  | REQUEST_CLIENT_CERT_NO_VERIFY | request_no_verify |
+  | REQUEST_CLIENT_CERT_AND_VERIFY | request_and_verify |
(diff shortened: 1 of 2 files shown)


ffoulkes:
Nuts. It appears that `rstcheck` doesn't work on correctly on files that use Sphinx macros.

Off the top of my head, this might be because:
- `rstcheck` isn't configured correctly
- It requires that additional packages be installed
- It doesn't work with Sphinx macros

I've removed `rstcheck` from the list of required checkers, to avoid blocking PRs until we have an answer.

docs/updates/development-update-24.07.md
```diff
@@ -0,0 +1,102 @@
...
+- The Stratum `CredentialsManager` class and `gnmi_cli` tool have been aligned
+  with the current upstream version, and IPDK modifications have been
+  overhauled to be more consistent with Stratum as a whole.
+
+  The `-grpc_client_cert_req_type` flag is now `-client_cert_req_type`, and
+  its value names have been shortened.
+
+  | Old Name | New Name |
+  | -------- | -------- |
+  | NO_REQUEST_CLIENT_CERT | NO_REQUEST |
```
5abeel, line 21: Should we update the new string names to lowecase based on the open PR?

5abeel:
Some candidates that we can add details on, based on the IPDK 24.01 release: https://github.com/ipdk-io/ipdk/blob/main/RELEASENOTES.md

* Update of Linux Networking program to the "v2" version
* Exception packet handlig
* Support for Link Aggregate Group (LAG)
* Packet IO
* Meters (think you have already mentioned these in this doc)
* Security enhancements
   - Upgrade from OpenSSL 1.1.1x to 3.x
   - Library updates to address security issues
           - Upgraded gRPC to 1.59.2 to address CVE-2023-33953, CVE-2023-4785
           - Upgraded protobuf to version 25.0
           - Upgraded abseil-cpp to version 20230802.0
           - Upgraded zlib to version 1.3
 
Upcoming attractions can include:
* Networking + IPsec combined recipe
* LAG in LACP mode
* Support for Geneve tunnel
* Default action support
* The "bf_" name changes?

ffoulkes:
The `bf_` name changes are specific to the E2100 SDE, and the SDE is not part of P4CP. We should only document aspects that are exposed through P4CP for the ES2K platform (whatever those may be).

docs/updates/development-update-24.07.md
ffoulkes: Done.

ffoulkes:
> Some candidates that we can add details on, based on the IPDK 24.01 release: https://github.com/ipdk-io/ipdk/blob/main/RELEASENOTES.md

I'm inclined to backfill with a 24.03 development update, rather than lump the 24.01 release into this one.
I'm also tempted to add a line stating that the development update corresponds to _&lt;IPDK or ES2100 release 24.01&gt;_.

5abeel:
> > Some candidates that we can add details on, based on the IPDK 24.01 release: https://github.com/ipdk-io/ipdk/blob/main/RELEASENOTES.md
> 
> I'm inclined to backfill with a 24.03 development update, rather than lump the 24.01 release into this one. I'm also tempted to add a line stating that the development update corresponds to _<IPDK or ES2100 release 24.01>_.

yes, we could do that. the dev update should contain everything since our last dev update which was 23.21 -- so, we could backfill with 24.03 instead of latest as of today.

PR #417, opened 2024-02-18
Move p4_role_config cmake build to stratum

ffoulkes:
- Moved the cmake code to build p4_role_config from networking-recipe to the stratum repository.

This is the first step toward relocating the remainder of the cmake build system for stratum from `networking-recipe` to the `stratum` repository. This should make it easier to make changes (especially adding new files), since we would only have to update one repository, instead of having to coordinate updates to both networking-recipe and stratum.

Paired with https://github.com/ipdk-io/stratum-dev/pull/206

### DIFF
--- a/stratum/CMakeLists.txt
+++ b/stratum/CMakeLists.txt
@@ -1,6 +1,6 @@
 # CMake build file for Stratum code.
 #
-# Copyright 2022-2023 Intel Corporation
+# Copyright 2022-2024 Intel Corporation
 # SPDX-License-Identifier: Apache 2.0
 #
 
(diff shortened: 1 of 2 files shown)


5abeel:
I had introduced a `target_link_libraries(ovs_sidecar_o PUBLIC stratum_static p4_role_config)` in ovs-p4rt client cmake file. Does this need to be updated with the proposed changes?

https://github.com/ipdk-io/networking-recipe/blob/main/ovs-p4rt/CMakeLists.txt#L70

5abeel:
May also want to update the stratum submodule before merging (or remove the upate from this PR, since submodule ref probably has been moved forwrad since)

ffoulkes:
> I had introduced a `target_link_libraries(ovs_sidecar_o PUBLIC stratum_static p4_role_config)` in ovs-p4rt client cmake file. Does this need to be updated with the proposed changes?
> 
> https://github.com/ipdk-io/networking-recipe/blob/main/ovs-p4rt/CMakeLists.txt#L70

Not a problem. Targets are defined globally and are visible throughout the cmake build system.

PR #418, opened 2024-02-19
Move cmake protobuf build logic from networking-recipe to stratum

ffoulkes:
Continue moving the cmake logic to build the stratum protobufs from `networking-recipe` to `stratum`.

The following protobuf builds have been relocated:

- common.proto
- error.proto
- p4_annotation.proto
- p4_control.proto
- p4_table_defs.proto

In addition to being moved into stratum itself, the protobuf builds are being disaggregated. Each protobuf is built individually and compiled into an object library. Explicit dependencies are used to ensure that the protobuf headers required to compile the source files are generated before the library is built. This gives us finer granularity, on par with what the Bazel build does.

I ran into problems when I tried to relocate the last half-dozen builds to the stratum repository, so this PR only does the disaggregation. Relocation will be done in a separate PR.

As part of the effort, `stratum/proto/CMakeLists.txt` has been divided into three separate files, one for each of the shared protobuf libraries we create (`libgrpc_proto.so`, `libp4runtime_proto.so`, and `libstratum_proto.so`). This makes the cmake build files easier to read and maintain.

I also fixed some load-time linkage issues with `libstratum_proto.so`.

Paired with https://github.com/ipdk-io/stratum-dev/pull/207

This PR will need to be rebased after https://github.com/ipdk-io/networking-recipe/pull/417 is merged.


5abeel (approved):
Same comment as other PR - submodule reference to stratum may need to be updated (or removed from this commit).

Rest LGTM

nupuruttarwar (approved):
Same comment as Sabeel. Also, please resolve conflicts before merging. Rest LGTM

PR #422, opened 2024-02-23
Move user guide source files to p4cp-userguide repository

ffoulkes:
This CL finishes moving the user guide source files to the https://ipdk-io/p4cp-userguide repository by deleting the old documents and updating the README file to provide a link to their new location.

The networking-recipe repository still has a number of Markdown files. so I haven't removed the `markdownlint` workflow.

There are currently no restructuredtext files, but someone might add one in the future, and we will want to catch any errors if that happens. I've retained the `rstdoc8` and `rstcheck` linters as well. The only change has been to use the non-sphinx version of `rstcheck`.

59 of the 61 "changed" files are deletions. The remaining two are what needs to be reviewed.


5abeel (approved):
LGTM

nupuruttarwar (approved):
LGTM

PR #427, opened 2024-03-02
Improve genscript help text

ffoulkes:
- Corrected name of script.

- Noted that template file must be in same directory as script.

- Parenthesized list of valid targets.


5abeel (approved):
LGTM

nupuruttarwar (approved):

PR #439, opened 2024-03-22
Fix condition for insecure mode check in sgnmi_cli

5abeel:
The flag that checks whether `sgnmi_cli` is using secure_mode or insecure_mode were inverted. Fixing the bug in this PR


5abeel:
> LGTM
> 
> Oops. (In retrospect, the flag should have been `--use_secure_mode` (or `--secure_mode`) with a default value of `True`. You generally want to assert a positive condition, not a negative condition.)

After I opened this PR, I had the same thought -- maybe we should move the block around so that it is an assert on positive condition. Let me refactor that since we are touching this section of the code now in this PR

PR #441, opened 2024-03-23
Add SDE patch scripts to repository

ffoulkes:
Added the DPDK and ES2K SDE patch scripts to the networking-recipe repository.

- The set-sde-target scripts add a TARGET file, which the Stratum Bazel buildsystem uses to determine which target the SDE supports.

- The patch-sde-rpaths scripts set the RUNPATH on several libraries, allowing the runtime loader to resolve transitive dependencies on other libraries without the assistance of LD_LIBRARY_PATH.

Also set execute permission on an unrelated Python script.

----

These scripts might reasonably go in an `sde` subdirectory, instead of the `dpdk` and `es2k` subdirectories.


5abeel:
Whats happening with the `scripts/common/extract_table_ids_from_p4info.py` file in this PR? GitHub is showing file modified and now an "empty file"??

ffoulkes:
> Whats happening with the `scripts/common/extract_table_ids_from_p4info.py` file in this PR? GitHub is showing file modified and now an "empty file"??

That's a peculiarity of GitHub. I changed the file permissions to include _execute_ (inasmuch as it's a Python script with a shebang, making it a _de facto_ executable). If you take a look to the right of the filename in the header, you'll see the change.

image

5abeel (approved):
LGTM

nupuruttarwar (approved):
LGTM

PR #446, opened 2024-04-05
Fix ES2K library directory list

ffoulkes:
The list of library directories for the ES2K SDE was incorrectly based on the SDE `include` directory instead of the base directory. This CL fixes the error.


nupuruttarwar (approved):
LGTM

PR #447, opened 2024-04-05
Define Linux Networking version for ovs-p4rt

ffoulkes:
Made the LNW_VERSION definitions available in networking-recipe, so the LNW_V2 and LNW_V3 conditionals are available for ovs-p4rt to use.


nupuruttarwar (approved):
LGTM

5abeel (approved):
LGTM

PR #456, opened 2024-04-20
Restructure ovs-p4rt

ffoulkes:
This is a fairly safe PR. It renames files and moves them to different locations (shown below), but -- except for editing a couple of #include file paths -- it doesn't make any changes to source code.

Most of what you'll see is that the main CMakeLists.txt file has been broken into five parts, separating the `ovs` section from the `sidecar` section and then dividing `ovs` into `testcontroller` and `vswitchd` sections.

----

1. Simplified filenames.
2. Modified buildsystem so the `dpdk` and `es2k` subdirectories add files directly to `ovs_sidecar_o`.
3. Divided `ovs-p4rt` into `ovs` and `sidecar` subdirectories.
4. Modified to use `OVS_INCLUDE_DIR` when building the sidecar.

----

The directory layout changes from this:
```text
ovs-p4rt/
|-- CMakeLists.txt
|-- cmake
|   `-- CheckBinutilsAVX512.cmake
|-- config.h.in
|-- dpdk
|   |-- CMakeLists.txt
|   `-- p4_name_mapping.h
|-- es2k
|   |-- CMakeLists.txt
|   `-- p4_name_mapping.h
|-- ovs_p4rt.cc
|-- ovs_p4rt_session.cc
|-- ovs_p4rt_session.h
|-- ovs_p4rt_tls_credentials.cc
`-- ovs_p4rt_tls_credentials.h
```
To this:
```text
ovs-p4rt/
|-- CMakeLists.txt
|-- cmake
|   `-- CheckBinutilsAVX512.cmake
|-- ovs
|   |-- CMakeLists.txt
|   |-- config.h.in
|   |-- testcontroller.cmake
|   `-- vswitchd.cmake
`-- sidecar
    |-- CMakeLists.txt
    |-- dpdk
    |   |-- CMakeLists.txt
    |   `-- p4_name_mapping.h
    |-- es2k
    |   |-- CMakeLists.txt
    |   `-- p4_name_mapping.h
    |-- ovsp4rt.cc
    |-- ovsp4rt_credentials.cc
    |-- ovsp4rt_credentials.h
    |-- ovsp4rt_session.cc
    `-- ovsp4rt_session.h
```

PR #458, opened 2024-04-22
Add logging to ovs-p4rt

ffoulkes:
This CL uses `glog` instead of `zlog` because ovs-p4rt already depends on Stratum code, which uses `glog`.

We can switch to using the Abseil logger when we upgrade to a newer version of Abseil.

This is an update to PR https://github.com/ipdk-io/networking-recipe/pull/456.

### DIFF
--- a/ovs-p4rt/sidecar/CMakeLists.txt
+++ b/ovs-p4rt/sidecar/CMakeLists.txt
@@ -33,5 +33,6 @@ add_dependencies(ovs_sidecar_o
 target_link_libraries(ovs_sidecar_o PUBLIC
     stratum_static
     p4_role_config_proto
+    absl::strings
 )
 
--- a/ovs-p4rt/sidecar/CMakeLists.txt
+++ b/ovs-p4rt/sidecar/CMakeLists.txt
@@ -10,6 +10,7 @@ add_library(ovs_sidecar_o OBJECT
     ovsp4rt_session.h
     ovsp4rt_credentials.cc
     ovsp4rt_credentials.h
+    ovsp4rt_logging.h
 )
 
 if(DPDK_TARGET)
--- a/ovs-p4rt/sidecar/ovsp4rt.cc
+++ b/ovs-p4rt/sidecar/ovsp4rt.cc
@@ -9,6 +9,7 @@
 #include "absl/flags/flag.h"
 #include "openvswitch/ovs-p4rt.h"
 #include "ovsp4rt_credentials.h"
+#include "ovsp4rt_logging.h"
 #include "ovsp4rt_session.h"
 
 #if defined(DPDK_TARGET)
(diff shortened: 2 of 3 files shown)


ffoulkes:
> Where do the logs go? Is a new file created in /var/log directory for the logs from this module? I dont see spec in this PR related to that.

Damn. I didn't think the problem through.

The log messages currently go to the console. They're roughly equivalent to the `printf`s they replace, so this is really no improvement.

We would normally be able to control the logging output through a myriad of command-line flags. These lags are only available if you use `gflags` to parse the command line. The sidecar is linked into `ovs-vswitchd` and `ovs-testcontroller`, which have their own command-line parsers, so we don't have access to the flags.

To use `glog` from the sidecar, I'd need to export a function that initializes the logger, and `ovs-vswitchd` and `ovs-testcontroller` would need to call the initialization on startup. I'd need to do the same thing in order to use `zlog`.

For `glog`, I'd have to provide a way for the user to express the logger configuration. `zlog` uses a `zlog-cfg` file, which `copy_config_files.sh` installs in `/usr/share/target_sys`. We've already defined an `OVSP4RT` output class. Which argues that my original solution (using the `target_sys` logger) was right after all.

ffoulkes:
Some additional ideas (which I'm going to share here, while they're fresh):

1. Have the sidecar use the ovs logger. This would couple it more tightly with ovs, but I don't think that's a problem.
2. Have the sidecar define an API that allows the caller to register a callback function. The sidecar calls the registered function to do logging.
3. Write a function that allows `glog` to be configured via a JSON file. Add a command-line flag to `ovs-vswitchd` and `ovs-testcontroller` that allows the user to specify the path to a glog configuration file. (Check to see if Abseil's new logger supports an external configuration file.)

ffoulkes:
Well, crap. 
`bf_log` turns out to be quite the litterbug.

Here's what you get on the terminal:
```text
dfoster@dgfoster-mobl1:~/work/latest$ ./build/ovs-p4rt/sidecar/log_msg_test
2024-04-25 13:05:03.119916 OVSP4RT ERROR - Error adding to FDB Tunnel Table: entry already exists
2024-04-25 13:05:03.120039 OVSP4RT ERROR - Error deleting entry from FDB Tunnel Table
2024-04-25 13:05:03.120101 OVSP4RT ERROR - Error adding entry to L2 Tunnel Table
2024-04-25 13:05:03.120161 OVSP4RT ERROR - Error deleting entry from FDB Source MAC Table
2024-04-25 13:05:03.120222 OVSP4RT ERROR - Error adding entry to FDB Rx Vlan Table
2024-04-25 13:05:03.120280 OVSP4RT ERROR - Error deleting entry from FDB Tx Vlan Table
2024-04-25 13:05:03.120337 OVSP4RT ERROR - Error adding entry to FDB Source MAC Table for b:e:e:b:e:e
2024-04-25 13:05:03.120392 OVSP4RT ERROR - Error deleting entry from SRC_IP_MAC_MAP_TABLE
2024-04-25 13:05:03.120451 OVSP4RT ERROR - Error adding entry to DST_IP_MAC_MAP_TABLE
```

Here's what it leaves behind in your working directory:
```text
dfoster@dgfoster-mobl1:~/work/latest$ ls -l *.log
-rw-r--r-- 1 dfoster dfoster   0 Apr 25 13:05 infrap4d.log
-rw-r--r-- 1 dfoster dfoster   0 Apr 25 13:05 krnlmon.log
-rw-r--r-- 1 dfoster dfoster 879 Apr 25 13:05 ovsp4rt.log
-rw-r--r-- 1 dfoster dfoster   0 Apr 25 13:05 p4_driver.log
-rw-r--r-- 1 dfoster dfoster   0 Apr 25 13:05 p4_sai.log
```

PR #460, opened 2024-04-22
Simplify generation of ovs build artifacts

ffoulkes:
- Defined `set_ovs_target_properties()` function to set properties that are common to both `ovs-vswitchd` and `ovs-testcontroller`.

This is an update to PR https://github.com/ipdk-io/networking-recipe/pull/456.


5abeel (approved):
LGTM

PR #461, opened 2024-04-25
Add logging to ovs-p4rt (bf_log)

ffoulkes:
This PR replaces the `glog` implementation (#458), which was deeply flawed. See the comments in the earlier PR for more information.

It is an update to PR #456.


ffoulkes:
Current take: the logger is poorly understood, and either its implementation or its application is deeply flawed.
- The mutable copy of the configuration file is in the working directory. We should be able to direct it to another location.
- The log files are in the working directory. We should be able to redirect them to another location (e.g. /var/log/infrap4d/).
- The default location of the config file is hard-wired (and there's a discrepancy between target-sys and target-utils).
- We should not be getting log files for classes/categories that don't apply to us.

PR #464, opened 2024-04-26
Implement --lnw-version parameter in make-all.sh

ffoulkes:
In make-all.sh:
- Parameters can be abbreviated to their shortest unique form, so `--lnw=2` is valid.
- Sets the `LNW_VERSION` cmake variable.
- If no value is specified, no setting is passed to cmake.

In the cmake buildsystem:
- `LNW_VERSION` is only processed when building the ES2K target.
- cmake validates the setting. Only `2` and `3` are permitted.
- Version defaults to `3` if `LNW_VERSION` is not specified.
- If specified, cmake:
  - Defines an `LNW_V2` or `LNW_V3` variable, for use within the buildsystem.
  - Passes `-DLNW_V2=1` or `-DLNW_V3=1` to the compiler when building krnlmon and ovs-p4rt.

### DIFF
--- a/make-all.sh
+++ b/make-all.sh
@@ -15,6 +15,7 @@ set -e
 
 _DEPS_DIR=${DEPEND_INSTALL}
 _HOST_DIR=${HOST_INSTALL}
+_LNW_VERSION=
 _NJOBS=8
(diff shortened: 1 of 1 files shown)


5abeel (approved):
LGTM

PR #465, opened 2024-04-28
Refine ovs-p4rt build

ffoulkes:
This PR makes it possible to do test builds of `ovs-p4rt` without having to build the entire networking-recipe.

- Moved the definition of `STRATUM_SOURCE_DIR` to the top-level listfile, to ensure that it is available to the ovs-p4rt and krnlmon builds.

- Updated the properties of the `ovs_sidecar_o` target to specify the dependencies needed to do a standalone build.

- Defined an on-demand `ovs_sidecar` target to generate a shared library, providing a concrete target that can be checked for unresolved external symbols.


5abeel (approved):
LGTM

PR #468, opened 2024-04-30
Move utility functions to ovsp4rt_utils

ffoulkes:
- Extracted utility functions from `ovsp4rt.cc` and moved them to `ovsp4rt_utils.cc`.
- Defined prototypes for the functions in `ovsp4rt_utils.h`.

### DIFF
--- a/ovs-p4rt/sidecar/common/ovsp4rt_utils.h
+++ b/ovs-p4rt/sidecar/common/ovsp4rt_utils.h
@@ -1,0 +1,44 @@
+// Copyright 2022-2024 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+
+#ifndef OVSP4RT_UTILS_H_
+#define OVSP4RT_UTILS_H_
+
+#include <netinet/in.h>
+
+#include <cstdint>
+#include <string>
+
+#include "p4/config/v1/p4info.pb.h"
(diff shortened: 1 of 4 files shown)


ovs-p4rt/sidecar/common/ovsp4rt_utils.h
```diff
@@ -0,0 +1,44 @@
...
+#include <netinet/in.h>
+
+#include <cstdint>
+#include <string>
+
+#include "p4/config/v1/p4info.pb.h"
+
+namespace ovs_p4rt {
+
+extern std::string CanonicalizeIp(const uint32_t ipv4addr);
```
5abeel, line 16: Declarations being introduced here. Did they not exist somewhere before or is there a cleanup needed someplace?

ffoulkes: 1. No function prototypes were required previously because the utility functions were in the same file as the functions that called them.
2. This PR moves the functions to a different source file, so that is no longer true.
3. We need a header file that defines prototypes for the utility functions, so the code in ovsp4rt.cc can continue to call them.
4. This PR defines a header file that defines the interface to the new file.

5abeel (approved):
LGTM

PR #469, opened 2024-04-30
ovsp4rt housekeeping

ffoulkes:
- Deleted unused definitions.

- Changed p4::v1 to ::p4::v1, #ifdef to #if defined(), and p4rt_grpc_addr to grpc_addr.

- Added #else/#error clauses to compound conditionals.

- Removed gratuitous return statements.


5abeel (approved):
LGTM

PR #470, opened 2024-05-01
Publish common function prototypes

ffoulkes:
- Created a header file for the private common functions, in preparation for refactoring ovsp4rt.cc into multiple source files. The prototypes for the public functions are already defined in ovs-p4rt.h.


5abeel (approved):
looks good

PR #471, opened 2024-05-01
Extract private DPDK functions

ffoulkes:
- Extracted private DPDK functions from ovsp4rt.cc and moved them to ovsp4rt_dpdk_private.cc. Created ovsp4rt_dpdk_private.h to provide function prototypes for the new module.


5abeel (approved):
LGTM

PR #472, opened 2024-05-01
Extract public DPDK functions

ffoulkes:
- Extracted public DPDK functions from ovsp4rt.cc and moved them to ovsp4rt_dpdk_public.cc.


5abeel (approved):
LGTM

PR #476, opened 2024-05-02
Extract public common functions (9/10)

ffoulkes:
- Extracted public common functions from ovsp4rt.cc and moved them to ovsp4rt_public.cc.


ovs-p4rt/sidecar/common/ovsp4rt_public.cc
```diff
@@ -0,0 +1,65 @@
...
+  status =
+      ConfigDecapTableEntry(session.get(), tunnel_info, p4info, insert_entry);
+  if (!status.ok()) return;
+#endif
+
+  status = ConfigTunnelTermTableEntry(session.get(), tunnel_info, p4info,
+                                      insert_entry);
+  if (!status.ok()) return;
+
+  return;
```
5abeel, line 64: Think its safe to remove this return statement, since it doesnt achieve anything

5abeel (approved):
One minor comment. Rest LGTM

ovs-p4rt/sidecar/common/ovsp4rt_public.cc
ffoulkes: So it is.

PR #480, opened 2024-05-06
Copy ovs-p4rt.h from ovs repository to ovs-p4rt

ffoulkes:
- Created a new `ovs-p4rt/include/ovsp4rt` directory. Copied the `ovs-p4rt.h` header file from the `ovs` repository to the `ovs-p4rt` include directory.

- Changed the #includes for the file from `"openswitch/ovs-p4rt.h"` to `"ovsp4rt/ovs-p4rt.h"`. Updated the cmake buildsystem to use the local include tree instead of the ovs include tree.

- In a separate commit (https://github.com/ipdk-io/ovs/pull/113), moved the copy of ovs-p4rt.h in the ovs repository from `include/openvswitch` to `include/ovsp4rt`, mirroring the layout of the ovs-p4rt include directory. Created a symlink from the old location to the new location, to maintain backward compatibility.

  The next step will be to modify the ovs buildsystem to accept the path of the ovs-p4rt include directory as a command-line parameter.

  We will continue to maintain two copies of `ovs-p4rt.h` for a reasonable period of time, to allow for a graceful transition, and then remove the copy from the ovs repository.

### DIFF
--- a/ovs-p4rt/CMakeLists.txt
+++ b/ovs-p4rt/CMakeLists.txt
@@ -6,5 +6,7 @@
 
 list(APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake)
 
+set(OVSP4RT_INCLUDE_DIR ${CMAKE_CURRENT_SOURCE_DIR}/include)
+
 add_subdirectory(sidecar)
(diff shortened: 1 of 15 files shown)


5abeel (approved):
LGTM

PR #425, opened 2024-02-26
Add support for combined recipe and linux networking version 3

nupuruttarwar:

PR #486, opened 2024-05-22
Address LNWv2 discrepancy in ovsp4rt

ffoulkes:
This CL addresses an apparent bug in PR https://github.com/ipdk-io/networking-recipe/pull/425.

The following unconditional change was made to common code as part of the LNWv3 implementation:

image

This PR restores the original code when networking-recipe is built for Linux Networking version 2.


ffoulkes:
Looking at the changes that were made in this function:

1. For V2, it used to set the action to L2_FWD\_**TX**\_TABLE_ACTION_L2_FWD. It is now setting it to L2_FWD\_**RX**\_TABLE_ACTION_L2_FWD.
2. For V2, it used to set the parameter value to learn_info.**src_port**. It not sets it to learn_info.**rx_src_port**, a value that did not exist in the original V2 code.

This means it's setting a different action and port number than it used to set. While this may be correct for the V3 P4 program, I don't see how it can be the correct behavior for the unmodified V2 program.

PR #490, opened 2024-06-06
Roll back ovs-p4rt refactoring

5abeel:
This reverts ovs-p4rt refactoring from b74d83d...892b1ea

Also reverts the ovs submodule changes introduced during those 10 commits


ffoulkes (approved):
This undoes four other commits in addition to the 10 you identified.

Also, you're not reverting ovs, you're reverting to an earlier version of ovs. (Which is a good thing, since I need the additional commits in order to upstream ovs.)

Approved

vsureshkumarp (approved):
LGTM

PR #491, opened 2024-06-06
Update krnlmon submodule ref

5abeel:
This PR updates the krnlmon reference to pick up the recent updates including enhancements and bug fixes


ffoulkes (commented):
It's generally a good idea to say why you're updating the submodule reference or what effect it will have. In this case, I believe the reason is "to pick up recent krnlmon updates."

The most important thing a comment does it to answer the question "why?"

vsureshkumarp (approved):
LGTM

ffoulkes (approved):
LGTM

PR #495, opened 2024-06-12
Build ovs-p4rt as a shared library

ffoulkes:
- Modified to build ovs-p4rt as a shared library instead of linking it statically into `ovs-vswitchd` and `ovs-testcontroller`.

- Modified build system to generate a `libovsp4rt.pc` pkg-config file.

----

Publishing ovs-p4rt as a shared library is a step toward making it a component in its own right.

Publishing ovsp4rt as a shared library should allow us to move generation of `ovs-vswitchd` and `ovs-testcontroller` back to the `ovs` repository, and will (I hope) make it possible for us to run the unit test suite against the P4 version of OvS. This is a key part of readying our P4 changes for upstreaming.


5abeel (approved):
lgtm

PR #498, opened 2024-06-13
Adding MAC addresses to logs when failure occurs

5abeel:
Adding MAC addresses in log printouts so that we can get more detailed information during debug sessions


ffoulkes (approved):
LGTM

PR #508, opened 2024-06-15
Reimplement basic logging for OVSP4RT

ffoulkes:
- Implemented a rudimentary logger for OVSP4RT, together with some helper functions to handle the most common message patterns. The initial implementation simply prints to the console.

- Replaced printf() calls with calls to the logger or helper functions.

### DIFF
--- a/ovs-p4rt/sidecar/lib/ovsp4rt_logging.cc
+++ b/ovs-p4rt/sidecar/lib/ovsp4rt_logging.cc
@@ -1,0 +1,35 @@
+// Copyright 2024 Intel Corporation.
+// SPDX-License-Identifier: Apache-2.0
+
+#include "ovsp4rt_logging.h"
+
+#include <stdarg.h>
+#include <stdio.h>
+
+namespace {
+
+const char* get_level_name(int level) {
+  switch (level) {
+    case OVSP4RT_LEVEL_ERROR:
+      return "ERROR";
+    case OVSP4RT_LEVEL_WARN:
+      return "WARN";
+    case OVSP4RT_LEVEL_INFO:
+      return "INFO";
+    case OVSP4RT_LEVEL_DEBUG:
+      return "DEBUG";
+    default:
+      return "UNKNOWN";
+  }
+}
+
+}  // namespace
+
+void ovsp4rt_log_message(int level, const char* format, ...) {
+  printf("OVSP4RT %s - ", get_level_name(level));
+  va_list args;
+  va_start(args, format);
+  vprintf(format, args);
+  va_end(args);
+  printf("\n");
+}
--- a/ovs-p4rt/sidecar/ovsp4rt.cc
+++ b/ovs-p4rt/sidecar/ovsp4rt.cc
@@ -6,6 +6,8 @@
 #include <string>
 
 #include "absl/flags/flag.h"
+#include "lib/ovsp4rt_logging.h"
+#include "lib/ovsp4rt_logutils.h"
 #include "ovsp4rt/ovs-p4rt.h"
 #include "ovsp4rt_credentials.h"
 #include "ovsp4rt_session.h"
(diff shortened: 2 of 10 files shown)


ovs-p4rt/sidecar/ovsp4rt.cc
```diff
@@ -2201,62 +2203,49 @@ void ovsp4rt_config_fdb_entry(struct mac_learning_info learn_info,
...
-    if (!status.ok())
-      printf(
-          "%s: Failed to program l2_fwd_tx_table for tunnel with mac "
-          "%x:%x:%x:%x:%x:%x\n",
-          insert_entry ? "ADD" : "DELETE", learn_info.mac_addr[0],
-          learn_info.mac_addr[1], learn_info.mac_addr[2],
-          learn_info.mac_addr[3], learn_info.mac_addr[4],
-          learn_info.mac_addr[5]);
+    if (!status.ok()) {
+      LogTableErrorWithMacAddr(insert_entry, "FDB Tunnel Table",
```
5abeel, line 2215: We should leave all table names as they are defined earlier since that is what we are looking at during debugging. Table names such as `l2_fwd_tx_table` or `l2_tunnel_to_v4_table` that can be correlated while also looking at the P4 program file

5abeel (changes requested):

ovs-p4rt/sidecar/lib/ovsp4rt_logging.cc
```diff
@@ -0,0 +1,35 @@
...
+      return "DEBUG";
+    default:
+      return "UNKNOWN";
+  }
+}
+
+}  // namespace
+
+void ovsp4rt_log_message(int level, const char* format, ...) {
+  printf("OVSP4RT %s - ", get_level_name(level));
```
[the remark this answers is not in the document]
ffoulkes: **>  is log file generation defined somewhere?**

This iteration is implemented using `printf()`. There's no log file; as noted in the CL comment, this version prints to the console.

The purpose of this CL is to:
- introduce the logger interface
- provide support functions to handle the most common message patterns (insert vs. delete; formatting MAC addresses)
- add logger calls in places that had a comment reading:
  ```c
  // TODO: print some log once logging support is added
  ```

This idea is to do the work iteratively, in a series of smaller commits.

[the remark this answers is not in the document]
ffoulkes: **> where is the default level defined?**

It isn't. Right now, everything's an error and everything gets printed. This maintains parity with the existing code.

The `printf`s don't provide any severity information, and I don't have the knowledge required to classify them. If you'd like to go through the code and assign levels, I'd be more than happy to replace `ovsp4rt_log_error()` to, say, `ovsp4rt_log_warn()`.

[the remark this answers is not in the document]
5abeel: ok, I think most of the errors printed in this module are error level, so they are appropriate for now

ovs-p4rt/sidecar/ovsp4rt.cc
ffoulkes: There are a number of problems with doing this:

1. It's not target/program-agnostic: it assumes that all P4 programs for all targets use the same table names for the same purposes, and will continue to do so in the future.
2. It bakes the table names into the code, in a section of the code that is separate from the places where the table names are defined and used.
3. It does not correlate with the software abstraction. The function `ConfigFdbTxVlanTableEntry()` maps to `L2_FWD_TX_TABLE`. 
4. Whom are the messages intended to benefit? The customer, or the developer?

Messages that are displayed by default (as these are) should be customer-friendly. This means using clear terminology and giving some indication of what the user is supposed to do in response to the message.

Messages that display P4 table names require inside knowledge. They are not customer friendly.

Error, warning, and info messages should be written with the customer in mind. Debug messages may be developer-focused.

I'm also concerned by the fact that the cavalier error handling. Failures during processing are sometimes logged and sometimes ignored. In most of the cases, processing continues after a failure has been detected. What state does this leave the remote system in? And you don't get a whole lot of information when you do get a log message. In some cases, there may several internal functions that update the same P4 table. Logging the table name with no additional information doesn't test you what kind of update failed. (That's one advantage of using the abstract name.) And except for the cases where a MAC address is displayed (something you recently expanded), there's no detail to help you understand which updates failed.

If I was participating in the design, I would request articulated philosophies/policies for both error handling and logging. And, if the policies didn't make it clear, I would ask for comments in the code explaining the decisions to act on some failures and ignore others.

"What we've got here is ... failure to communicate." _Cool Hand Luke_

ffoulkes: In any case, I was thinking of creating a bunch of `constexpr`s to define the text to be displayed, labeled with the abstract name, and putting them in a header file. This would handle any differences between DPDK and ES2K, and it would make it to change our minds over what to display.

ffoulkes: I might also create a DETAIL message class that can be used to enable additional information to be displayed with a DEBUG message.

ffoulkes: Please mull these thoughts over. We can then talk about how best to proceed.

ffoulkes:
I've reengineered the logging code to display table names without violating the abstraction model.

ovs-p4rt/sidecar/ovsp4rt.cc
ffoulkes: I've reengineered the logging code to display table names without violating the abstraction model.

5abeel (approved):
LGTM

PR #509, opened 2024-06-16
Add support for OVSP4RT unit testing

ffoulkes:
- Created infrastructure for ovs-p4rt unit testing.
  - Implemented a `stratum_utils` library, which provides the facilities used by ovs-p4rt.
    This makes it possible to build ovs-p4rt without building all of Stratum.
  - Created a header file (`ovsp4rt_private.h`) to expose prototypes for the functions under test.
- Developed unit test for `PrepareL2ToTunnelV4()` and `PrepareL2ToTunnelV6()`.
- Eliminated unnecessary dependencies from build.


ovs-p4rt/sidecar/testing/prepare_l2_to_tunnel_test.cc
```diff
@@ -0,0 +1,198 @@
...
+  // Act
+  ::p4::v1::TableEntry table_entry;
+  DiagDetail detail;
+  PrepareL2ToTunnelV6(&table_entry, fdb_info, p4info, true, detail);
+#ifdef DUMP_JSON
+  DumpMessageAsJson(table_entry);
+#endif
+
+  // Assert
+  ASSERT_EQ(table_entry.table_id(), kTableIdV6);
```
n-sandeep, line 160: Should this PR also include a p4info.txt file?
which has kTableIdV4, kActionIdV4, kTableIdV6 and kActionIdV6

ovs-p4rt/sidecar/testing/prepare_l2_to_tunnel_test.cc
```diff
@@ -0,0 +1,198 @@
...
+  auto action = table_action.action();
+  ASSERT_EQ(action.action_id(), kActionIdV4);
+
+  ASSERT_EQ(action.params_size(), 1);
+
+  auto param = action.params()[0];
+  ASSERT_EQ(param.param_id(), 1);
+
+  auto param_value = param.value();
+  ASSERT_EQ(param_value.size(), 4);
```
n-sandeep, line 121: I am not very sure about this value.
but looking at sample p4info.txt for LNWv3, i can see IPv4 has one param only

actions {
  preamble {
    id: 23805991
    name: "linux_networking_control.set_tunnel_v4"
    alias: "set_tunnel_v4"
  }
  **params {**
    id: 1
    name: "dst_addr"
    bitwidth: 32
  }
}

But same IPv6 has 4 params.
actions {
  preamble {
    id: 23953453
    name: "linux_networking_control.set_tunnel_v6"
    alias: "set_tunnel_v6"
  }
  params {
    id: 1
    name: "ipv6_1"
    bitwidth: 32
  }
  params {
    id: 2
    name: "ipv6_2"
    bitwidth: 32
  }
  params {
    id: 3
    name: "ipv6_3"
    bitwidth: 32
  }
  params {
    id: 4
    name: "ipv6_4"
    bitwidth: 32
  }
}

n-sandeep, line 121: not sure about this value, can you please let us know what does this param_value size point to? is it size of IPv4 address in bytes?

ovs-p4rt/sidecar/testing/prepare_l2_to_tunnel_test.cc
ffoulkes: I'd like to include it. I was concerned that it might be considered proprietary.

ovs-p4rt/sidecar/testing/prepare_l2_to_tunnel_test.cc
ffoulkes: `FieldMatch.exact.value` is a `bytes` field. `bytes` fields are represented in C++ as a `std::string`. `size()` returns the length of the string. The assertion verifies that the size of the `value` field in the the `TableEntry` object is of the expected size.

n-sandeep (approved):
Changes LGTM

ovs-p4rt/sidecar/testing/p4Info.txt
```diff
@@ -0,0 +1,2760 @@
+pkg_info {
```
5abeel, line 1: Since this file is a compiler-generated artifact, this file will need to get periodically updated with new compiler versions (frequency of compiler version change is 2-3x per month). 

It would be a good idea to add some metadata to indicate which p4c compiler generated this file (ideally outside of this file, so we can simply drop the p4Info.txt file without editing it).

5abeel: Another change that could trigger this is the p4 program file change itself. I am not aware of versioning of individual p4 programs - perhaps that not available (but a change will trigger a new p4Info from p4c downstream from those changes)

ovs-p4rt/sidecar/testing/scenarios.md
```diff
@@ -0,0 +1,98 @@
...
+  - ip_info.dst_ip_addr.ip.family := AF_INTE
+  - ip_info.dst_ip_addr.ip.v4addr.s_addr := _dst_
+- [ovsp4rt_config_ip_mac_map_entry](#ovsp4rt_config_ip_mac_map_entry)
+  - insert_entry = false
+
+### update_ip_mac_map_info
+
+(to be documented)
+
+### xlate_normal
```
5abeel, line 58: Translate? Should just write translate instead to keep the messaging clear

5abeel (approved):
Couple of comments on suggested improvement. Rest LGTM

ovs-p4rt/sidecar/testing/p4Info.txt
ffoulkes: We don't need to update the p4Info file every time there's a change in the compiler output. In fact, it's probably in our best interests to continue using the same version of the file as long as we can. We're looking for unintentional changes in behavior between two versions of the _code_.

This doesn't mean that we shouldn't have the capability to replace the p4Info file and automatically update the data the tests are sensitive too. I figured we'd write a program that reads p4Info.txt, extracts the values it needs, and writes a bunch of `constexpr`s to a header file that the test programs can `#include`.

PR #510, opened 2024-06-19
Fix l2_to_tunnel_v6 table delete

n-sandeep:
When a MAC delete happens, we will not have info regarding this MAC belongs to an Ipv4/IPv6 client. Also, we will not know if this MAC is learned on tunnel interface or regular interface.
On a FDB delete, we check l2_to_tunnel_v4 and l2_to_tunnel_v6 tables to see if this is a v4 or v6 tunnel.
If V6 tunnel, we should also set AF_INET6 flags for both local and remote IP as this info is used in ConfigL2TunnelTableEntry API. Earlier this AF_INET6 family type was not set, so even for IPv6 entry we try to delete from l2_to_tunnel_v4 instead of l2_to_tunnel_v6 table.


ffoulkes (approved):
LGTM

PR #512, opened 2024-06-24
Rename the 'lib' directory to 'logging'

ffoulkes:
Given that the only files in the `lib` directory are related to logging, and there are six of them (seven with CMakeLists.txt), it makes sense to make the directory logging-specific.

Not urgent, but the sooner this is merged, the better.


5abeel (approved):
LGTM

PR #518, opened 2024-06-27
Update script files

ffoulkes:
- Addressed shellcheck issues.

- Revised `copy_config-files.sh`.
  - Specified `-v` on `cp` commands to provide feedback.
  - Replaced `return 0` (invalid outside a function) with `exit 0` (indicating that the script failed).


5abeel (approved):
LGTM

PR #520, opened 2024-07-01
Define ovsp4rt_spies library

ffoulkes:
- Created a library of test doubles (spies) for the OVSP4RT API that capture input parameters and export them as JSON.

- Implemented a spy for the ovsp4rt_config_fdb_entry API.

- Created a manual test for the spy.

- Moved ovsp4rt_logging_test to the 'logging' directory.

- Moved the pkg-config templates to a new 'cmake' directory.

----

### Notes

1. The plan is to use the spies to capture test cases that can be used to create approval (characterization) tests for the legacy OVSP4RT code. Michael Feathers recommends using characterization tests to check for regressions when refactoring legacy code. (_Working Effectively with Legacy Code_) 

   With approval tests, you capture sets of inputs and outputs for each unit before you refactor it. Your test invokes the unit with a set of inputs, captures the outputs, and verifies them against the outputs you captured when you first characterized the unit. You run the tests repeatedly as you refactor, checking whether your (supposedly harmless) changes have caused any of the outputs to change.

   It is difficult to impossible to write unit tests for a legacy system. Approval testing is a good alternative.

2. This PR adds new files and moves a couple of existing files to new directories. The only modifications are to the cmake build system.

3. I'm using Niels Lohmann's [JSON for Modern C++](https://json.nlohmann.me/). This is the same library that Google uses in Protobuf. It's part of stratum-deps.

4. I expect the code to be refactored frequently as the library evolves. Among other things, I expect to extract the code that renders input data structures as JSON into a set of utility functions.

----

### Plans

I'm planning to do two things:

1. Focus on creating approval tests, rather than unit tests, for OVSP4RT.
2. Take another shot at refactoring OVSP4RT, using the approval tests to verify that I haven't altered the behavior of the code.

To do this, I need to:

1. Create a set of functions that serialize the inputs to each of the public APIs.
2. Instrument OVSP4RT to capture the inputs and outputs of each of the public API functions. The outputs are Protobufs, so conversion to JSON should be straightforward.
3. Create a set of test cases in OVS. Use the instrumentation to capture the corresponding transactions in OVSP4RT.
4. Create a suite of approval tests that can feed the inputs into OVSP4RT, capture the outputs, and check for deviations from the expected results.
5. Except for instrumenting the input and output paths (which I plan to make a compile option), the majority of the updates will be orthogonal to OVSP4RT itself.

I started with `ovsp4rt_config_fdb_entry()` because that's the API for which I wrote the unit test. I used it to learn how to serialize a set of inputs into a JSON object. The next step is to instrument its input and output paths, and then rejigger the unit test to create a corresponding approval test.

I suspect the ovsp4rt_spies library will morph into a support library for the instrumented version of OVSP4RT as well as the approval tests. And I've been wondering whether I might be able to use it in conjunction with some of the programs in the OVS test suite...


ovs-p4rt/sidecar/spies/ovsp4rt_config_fdb_entry_spy.cc
```diff
@@ -0,0 +1,75 @@
...
+
+#include <stdbool.h>
+#include <stdint.h>
+
+#include <iostream>
+#include <nlohmann/json.hpp>
+
+#include "ovsp4rt/ovs-p4rt.h"
+
+void ovsp4rt_config_fdb_entry(struct mac_learning_info learn_info,
```
5abeel, line 17: I read the notes and description on the main page of PR.

I still dont think I follow what the intention is - are we planning a refactor of functions such as `ovsp4rt_config_fdb_entry`?

ffoulkes: I'm planning to do two things:

1. Focus on creating approval tests, rather than unit tests, for OVSP4RT.
2. Take another shot at refactoring OVSP4RT, using the approval tests to verify that I haven't altered the behavior of the code.

To do this, I need to:

1. Create a set of functions that serialize the inputs to each of the public APIs.
2. Instrument OVSP4RT to capture the inputs and outputs of each of the public API functions.
3. Create a set of test cases in OVS. Use the instrumentation to capture the corresponding transactions in OVSP4RT.
4. Create a suite of approval tests that can feed the inputs into OVSP4RT, capture the outputs, and check for deviations from the expected results.

I started with `ovsp4rt_config_fdb_entry()` because that's the API for which I wrote the unit test. I used it to learn how to serialize a set of inputs into a json object. The next step is to instrument its input and output paths, and then rejigger the unit test to create a corresponding approval test.

I suspect the `ovsp4rt_spies` library will morph into a support library for the instrumented version of OVSP4RT as well as the approval tests. And I've been wondering whether I might be able to use it in conjunction with some of the programs in the OVS test suite...

5abeel: I am ok with creating approval tests and have no issues with that.

With regards to refactoring, please check with @vsureshkumarp and @n-sandeep on the repurcussions. Suresh was suspecting that the last round of refactoring may have messed with the timing mechanism and the flow (since each individual method itself seemed ok), which caused the errors/issues.

5abeel: Approving this PR, since there are no functional changes

5abeel (approved):
lgtm

ovs-p4rt/sidecar/spies/ovsp4rt_config_fdb_entry_spy.cc
ffoulkes: > With regards to refactoring, please check with @vsureshkumarp and @n-sandeep on the repercussions. Suresh was suspecting that the last round of refactoring may have messed with the timing mechanism and the flow (since each individual method itself seemed ok), which caused the errors/issues.

That seems unlikely. The last round of refactoring is gone.

PR #530, opened 2024-07-08
Fix VNI issue of setting only 1 byte

5abeel:
A bug was reported that VNI values were not programmed correctly and only 1 byte was copied over. This fixes the reported issue for ES2K target

### DIFF
--- a/ovs-p4rt/sidecar/ovsp4rt.cc
+++ b/ovs-p4rt/sidecar/ovsp4rt.cc
@@ -336,7 +336,10 @@ void PrepareFdbTableEntryforV4VxlanTunnel(
           param->set_param_id(GetParamId(
               p4info, L2_FWD_TX_TABLE_ACTION_POP_VLAN_SET_VXLAN_UNDERLAY_V4,
               ACTION_PARAM_TUNNEL_ID));
-          param->set_value(EncodeByteValue(1, learn_info.tnl_info.vni));
+          param->set_value(
+              EncodeByteValue(3, (learn_info.tnl_info.vni >> 16) & 0xFF,
+                              (learn_info.tnl_info.vni >> 8) & 0xFF,
+                              learn_info.tnl_info.vni & 0xFF));
         }
       } else {
         action->set_action_id(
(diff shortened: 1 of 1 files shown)


ovs-p4rt/sidecar/ovsp4rt.cc
```diff
@@ -336,7 +336,10 @@ void PrepareFdbTableEntryforV4VxlanTunnel(
           param->set_param_id(GetParamId(
               p4info, L2_FWD_TX_TABLE_ACTION_POP_VLAN_SET_VXLAN_UNDERLAY_V4,
               ACTION_PARAM_TUNNEL_ID));
-          param->set_value(EncodeByteValue(1, learn_info.tnl_info.vni));
+          param->set_value(
```
ffoulkes, line 339: 1. I would recommend that, instead of repeating the procedure to encode a vni 39 times, you create an `EncodeVniValue()` helper function and use that throughout. This would then become:
   ```c
   param->set_value(EncodeVniValue(learn_info.tnl_info.vni));
   ```
   which is easier to read and maintain.
2. Speaking of which: `learn_info.tnl_info.vni` is a `uint16_t`. It's 2 bytes wide, not 3. I would therefore propose:
   ```c
   inline std::string EncodeVniValue(uint16_t vni) {
     return EncodeByteValue(2, (vni >> 8) & 0xFF, vni & 0xFF);
   }
   ```
3. If, for some unfathomable reason, the encoded two-byte value really needs to be three bytes wide, I would suggest:
   ```c
   inline std::string EncodeVniValue(uint16_t vni) {
     return EncodeByteValue(3, 0, (vni >> 8) & 0xFF, vni & 0xFF);
   }
   ```
   which is clearer than shifting an unsigned 16 bit value right 16 bits (guaranteed to yield 0) and then ANDing it with 0xff (also guaranteed to yield 0).

ffoulkes (commented):
1. I would recommend that, instead of repeating the procedure to encode a vni 39 times, you create an `EncodeVniValue()` helper function and use that throughout. This would then become:
   ```c
   param->set_value(EncodeVniValue(learn_info.tnl_info.vni));
   ```
   which is easier to read and maintain.
2. Speaking of which: `learn_info.tnl_info.vni` is a `uint16_t`. It's 2 bytes wide, not 3. I would therefore propose:
   ```c
   static inline std::string EncodeVniValue(uint16_t vni) {
     return EncodeByteValue(2, (vni >> 8) & 0xFF, vni & 0xFF);
   }
   ```
3. If, for some unfathomable reason, the encoded two-byte value really needs to be three bytes wide, I would suggest:
   ```c
   static inline std::string EncodeVniValue(uint16_t vni) {
     return EncodeByteValue(3, 0, (vni >> 8) & 0xFF, vni & 0xFF);
   }
   ```
   which is clearer than shifting an unsigned 16 bit value right 16 bits (guaranteed to yield 0) and then ANDing it with 0xff (also guaranteed to yield 0).

5abeel:
Addressed comments. Much easier to read and follow/compare to previous version.

ffoulkes (approved):
A considerable improvement. Approved.

PR #532, opened 2024-07-09
Modify prepare_l2_to_tunnel_test to support --dump_json flag

ffoulkes:
- Implemented `test_main.cc`, which allows googletest programs to define Abseil flags.
- Modified `prepare_l2_to_tunnel_test` to dump output under control of the `--dump_json` command-line flag.


5abeel (approved):
LGTM

PR #533, opened 2024-07-09
Begin development of Tracker class

ffoulkes:
### Synopsis

The `Tracker` class is designed to capture the inputs and outputs of a call to the ovs-p4rt API, for use in testing. It is currently under development.

### Usage

The Tracker object is instantiated by the public API function, which uses it to capture its input parameters.

For example:

```c
Tracker tracker;
tracker.captureInput(__func__, learn_info, insert_entry);
```

The public function passes a reference to the object to the internal functions it calls. Prior to sending a WriteRequest, the internal function captures the output:

```c
tracker.captureOutput(write_request);
```

At the end of processing the API function saves the tracked parameters for subsequent use:

```c
tracker.saveSnapshot();
```

### Impact

This PR is orthogonal to the rest of ovs-p4rt. The only connection is an `add_subdirectory(tracker)` in `sidecar/CMakeLists.txt`, which is enabled by the `BUILD_TRACKER` option.

### DIFF
--- a/ovs-p4rt/sidecar/CMakeLists.txt
+++ b/ovs-p4rt/sidecar/CMakeLists.txt
@@ -4,6 +4,8 @@
 # SPDX-License-Identifier: Apache 2.0
 #
 
+option(BUILD_TRACKER "Build Tracker library (under development)" ON)
+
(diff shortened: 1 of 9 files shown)


5abeel (approved):
lgtm

PR #537, opened 2024-07-11
Encode p4info file

ffoulkes:


ovs-p4rt/sidecar/testing/encode_host_port_value_test.cc
```diff
@@ -0,0 +1,66 @@
...
+
+#include <string>
+
+#include "gtest/gtest.h"
+#include "ovsp4rt/ovs-p4rt.h"
+
+//----------------------------------------------------------------------
+// Support function (from ovsp4rt.cc)
+//----------------------------------------------------------------------
+std::string EncodeByteValue(int arg_count...) {
```
5abeel, line 15: Shouldnt we be using the actual function from the ovsp4rt.cc instead of a repeated/mock function which duplicates the code?

ovs-p4rt/sidecar/testing/p4info_text.h
```diff
@@ -0,0 +1,2770 @@
+// Copyright 2024 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+
+#ifndef TESTING_P4INFO_TEXT_H_
+#define TESTING_P4INFO_TEXT_H_
+
+// P4Info for ES2K Linux Networking V3.
+constexpr char P4INFO_TEXT[] =
```
5abeel, line 8: Good call on uses the text as .h file instead of disk reads.

ovs-p4rt/sidecar/testing/encode_host_port_value_test.cc
ffoulkes: It depends. Sometimes you're independently verifying something, and don't want to rely on the implementation in the function used by the Unit Under Test. I don't think that's the case here, but I'll need to double-check the code to be sure.

Good suggestion.

ovs-p4rt/sidecar/testing/p4info_text.h
ffoulkes: Thanks. 😊

ovs-p4rt/sidecar/testing/encode_host_port_value_test.cc
ffoulkes: On review, `EncodeByteValue()` is used to establish the control value. It should be the actual implementation, not a private copy. You're right.

I've implemented your suggestion. Good call. 👍

5abeel (approved):
LGTM

PR #541, opened 2024-07-11
Move set_install_rpath() to its own file

ffoulkes:
- Moved the definition of the `set_install_rpath()` function to `SetInstallRpath.cmake` so it can be used in the standalone krnlmon build.

This PR is associated with https://github.com/ipdk-io/krnlmon/pull/154.


5abeel (approved):
LGTM

PR #542, opened 2024-07-11
Revise src_port verification

ffoulkes:
The `src_port` field contains an arbitrary value that has nothing to do with what was specified in the tunnel_info structure. This is a workaround for a long-standing bug in the Linux Networking P4 program.

Documented this fact in the test program and placed the check under control of a `--check_src_port` command-line flag that defaults to False.

### DIFF
--- a/ovs-p4rt/sidecar/testing/geneve_encap_table_entry_test.cc
+++ b/ovs-p4rt/sidecar/testing/geneve_encap_table_entry_test.cc
@@ -137,7 +137,6 @@ TEST_F(GeneveEncapTableEntryTest, encap_param_src_port_is_correct) {
     }
   }
 
-
   if (absl::GetFlag(FLAGS_check_src_port)) {
     // The src_port field contains an arbitrary value that has nothing
     // to do with what was specified in the tunnel_info structure.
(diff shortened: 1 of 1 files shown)


ovs-p4rt/sidecar/testing/geneve_encap_table_entry_test.cc
```diff
@@ -136,10 +137,14 @@ TEST_F(GeneveEncapTableEntryTest, encap_param_src_port_is_correct) {
     }
   }
 
-  // PrepareGeneveEncapTableEntry sets the dst_port parameter twice.
-  // The src_port parameter is not set at all. We get garbage instead.
-  ASSERT_TRUE(src_port.has_value());
-  // ASSERT_EQ(src_port.value(), SRC_PORT);
+  if (absl::GetFlag(FLAGS_check_src_port)) {
+    // The src_port field contains an arbitrary value that has nothing
```
5abeel, line 141: Perhaps we should add this comment to the actual source code file too in ovsp4rt.cc so that its evident that this is an intentional use by us.

5abeel (approved):
LGTM

ovs-p4rt/sidecar/testing/geneve_encap_table_entry_test.cc
ffoulkes: I agree.

PR #544, opened 2024-07-11
Annotate members with P4 bit widths

ffoulkes:
- Annotated struct members with the width in bits of the underlying field in the P4 program.


5abeel (approved):
lgtm

PR #557, opened 2024-07-15
Update submodules

5abeel:
This PR updates the submodule references for krlmon and stratum repos.

The krnlmon ref update includes LAG use-case bug fixes. stratum update includes updating the es2k_skip_p4.config file


ffoulkes (approved):
LGTM

PR #562, opened 2024-07-16
[ovsp4rt] Revise IPv4 and IPv6 test values

ffoulkes:
- Updated IPV4_SRC_ADDR so each octet is unique.

- Updated SRC_PORT, DST_PORT, and VNI so all values are 16 bits and each octet is unique.

This should help detect byte swap errors.


5abeel (approved):
LGTM

PR #565, opened 2024-07-16
[ovsp4rt] Make the spies library optional

ffoulkes:
- Placed generation of the spies library under control of a BUILD_SPIES option (which is OFF by default).

This feature is still in the early stages of development. It should be disabled by default.


5abeel (approved):
lgtm

PR #566, opened 2024-07-17
[ovsp4rt] Improve isolation of spies library build

ffoulkes:
- Moved the logic to install the spies library, and to build and and install the pkg-config file, to spies/CMakeLists.txt.

- Fixed a typo in the library install() command.


5abeel (approved):
lgtm

PR #571, opened 2024-07-18
Define ovsp4rt-unit-tests build target

ffoulkes:
- Implemented an `ovs-ovsp4rt-tests` target to build all the ovsp4rt unit tests.

- Modified `vxlan_encap_v4_table_entry_test` to check the src_port only when built for the ES2K target. The DPDK version of `PrepareVxlanEncapTableEntry()` does not encode this parameter.


5abeel (approved):
LGTM

PR #579, opened 2024-07-23
[ovsp4rt] Fix byte-order bug in EncodeVniValue()

ffoulkes:
- Corrected a byte-order but in EncodeVniValue().

- Fixed error InitV4TunnelInfo() and InitV6TunnelInfo(). The src_addr, dst_addr, and vni fields should be in host byte order.

- Implemented DecodePortValue() and DecodeVniValue() methods to make it easier to change algorithms in the future.

### DIFF
--- a/ovs-p4rt/sidecar/ovsp4rt.cc
+++ b/ovs-p4rt/sidecar/ovsp4rt.cc
@@ -48,6 +48,10 @@ std::string EncodeByteValue(int arg_count...) {
   return byte_value;
 }
 
+static inline std::string EncodeVniValue(uint16_t vni) {
+  return EncodeByteValue(3, 0, (vni >> 8) & 0xFF, vni & 0xFF);
+}
+
 std::string CanonicalizeIp(const uint32_t ipv4addr) {
   return EncodeByteValue(4, (ipv4addr & 0xff), ((ipv4addr >> 8) & 0xff),
                          ((ipv4addr >> 16) & 0xff), ((ipv4addr >> 24) & 0xff));
@@ -113,10 +117,6 @@ int GetMatchFieldId(const ::p4::config::v1::P4Info& p4info,
 static inline int32_t ValidIpAddr(uint32_t nw_addr) {
   return (nw_addr && nw_addr != INADDR_ANY && nw_addr != INADDR_LOOPBACK &&
           nw_addr != 0xffffffff);
-}
-
-static inline std::string EncodeVniValue(uint16_t vni) {
-  return EncodeByteValue(3, 0, vni & 0xFF, (vni >> 8) & 0xFF);
 }
 
 #if defined(ES2K_TARGET)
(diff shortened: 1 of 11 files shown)


ovs-p4rt/sidecar/testing/ipv4_tunnel_test.h
```diff
@@ -17,7 +17,7 @@ constexpr char IPV4_DST_ADDR[] = "192.168.17.5";
 constexpr int IPV4_PREFIX_LEN = 24;
 
 constexpr uint16_t SRC_PORT = 0x1066;
-constexpr uint16_t DST_PORT = 0x1984;
+constexpr uint16_t DST_PORT = 0x4224;
```
ffoulkes, line 20: 0x1984 overflows when you swap the upper and lower bytes (0x8419) and multiply by two. Replaced it with a value that can be doubled without overflowing.

ovs-p4rt/sidecar/testing/ipv4_tunnel_test.h
```diff
@@ -35,9 +35,9 @@ class Ipv4TunnelTest : public TableEntryTest {
         << "Error converting " << IPV4_DST_ADDR;
     info.remote_ip.prefix_len = IPV4_PREFIX_LEN;
 
-    info.src_port = htons(SRC_PORT);
-    info.dst_port = htons(DST_PORT);
-    info.vni = htons(VNI);
+    info.src_port = SRC_PORT;
```
ffoulkes, line 38: The `info` struct is what's normally passed in by OVS. Fields are assumed to be in native byte order.

ovs-p4rt/sidecar/testing/table_entry_test.h
```diff
@@ -42,6 +42,17 @@ class TableEntryTest : public ::testing::Test {
     }
   }
 
+  static uint16_t DecodePortValue(const std::string& string_value) {
```
ffoulkes, line 45: Abstracted the specific decoding algorithms for Port and VNI parameters, reducing repetition and making it easier to make changes to those algorithms if the need arises (as was the case with this PR).

ovs-p4rt/sidecar/ovsp4rt.cc
```diff
@@ -48,6 +48,10 @@ std::string EncodeByteValue(int arg_count...) {
   return byte_value;
 }
 
+static inline std::string EncodeVniValue(uint16_t vni) {
+  return EncodeByteValue(3, 0, (vni >> 8) & 0xFF, vni & 0xFF);
```
5abeel, line 52: So, this restores to previous encoding byte order.

We will still have the issue of IPv6 not working for over 1024 values, but thats a smaller problem than what we have currently. We can take the IPv6 issue on later

5abeel (approved):
LGTM

ovs-p4rt/sidecar/ovsp4rt.cc
ffoulkes: Small batches. Incremental improvement. It's one of the core principles of modern software development.

PR #585, opened 2024-07-26
OBSOLETE: [ovsp4rt] Support 24-bit VNI values

ffoulkes:
- Widened `tunnel_info.vni` from uint16_t to uint32_t.

- Modified `EncodeVniValue()` to encode the low-order 24 bits of a uint32_t parameter.

- Implemented a new `EncodeTunnelId()` function to use in place of `EncodeVniValue()` when the destination is a tunnel_id field. It encodes the low-order 20 bits of a uint32_t parameter. _Note that these VNI values are 20 bits, not 24._

- Fixed a bug in `PrepareFdbTableEntryforV4VxlanTunnel()`, which encoded VNI as 8 bits for an IPv6 tunnel when the port_vlan_mode is P4_PORT_VLAN_NATIVE_UNTAGGED.

- Modified unit tests to use uint32_t instead of uint16_t to represent VNI values.

- Changed IPv4 VNI test value from 16 to 24 bits.

Paired with https://github.com/ipdk-io/ovs/pull/132.


ovs-p4rt/sidecar/testing/ipv4_tunnel_test.h
```diff
@@ -18,7 +18,7 @@ constexpr int IPV4_PREFIX_LEN = 24;
 
 constexpr uint16_t SRC_PORT = 0x1066;
 constexpr uint16_t DST_PORT = 0x4224;
-constexpr uint16_t VNI = 0x1776;
+constexpr uint32_t VNI = 0x177607;
```
ffoulkes, line 21: Make sure one of the test values is 24-bit.

ovs-p4rt/sidecar/ovsp4rt.cc
```diff
@@ -375,8 +381,8 @@ void PrepareFdbTableEntryforV4VxlanTunnel(
           param->set_param_id(GetParamId(
               p4info, L2_FWD_TX_TABLE_ACTION_POP_VLAN_SET_VXLAN_UNDERLAY_V6,
               ACTION_PARAM_TUNNEL_ID));
-          // TODO(derek): 8-bit vni (es2k, ipv6)
-          param->set_value(EncodeByteValue(1, learn_info.tnl_info.vni));
+          // tunnel_id is bit<20>
+          param->set_value(EncodeTunnelId(learn_info.tnl_info.vni));
```
ffoulkes, line 385: Bug fix

ovs-p4rt/sidecar/ovsp4rt.cc
```diff
@@ -352,7 +358,7 @@ void PrepareFdbTableEntryforV4VxlanTunnel(
           param->set_param_id(GetParamId(
               p4info, L2_FWD_TX_TABLE_ACTION_POP_VLAN_SET_VXLAN_UNDERLAY_V4,
               ACTION_PARAM_TUNNEL_ID));
-          param->set_value(EncodeVniValue(learn_info.tnl_info.vni));
+          param->set_value(EncodeTunnelId(learn_info.tnl_info.vni));
```
5abeel, line 361: Why change it from `EncodeVniValue` to `EncodeTunnelId`?? The member variable that is part of the name defined in `struct tunnel_info` is called `vni`. 

It would be better and cleaner to keep the name consistent instead of introducing new naming

ovs-p4rt/include/ovsp4rt/ovs-p4rt.h
```diff
@@ -86,7 +86,7 @@ struct tunnel_info {
   struct p4_ipaddr local_ip;
   struct p4_ipaddr remote_ip;
   uint16_t dst_port;  // bit<16>
-  uint16_t vni;       // bit<24>
+  uint32_t vni;       // bit<24>
```
5abeel, line 89: What are the repurcussions of changing this to 32-bit here and not in OVS codebase? 

I understand that there is a copy of the ovs-p4rt.h in OVS repo, which still has the uint16.

Further, even if we update that OVS copy to uin32 - need to make sure that OVS can use the corrected value and all references are checked

ovs-p4rt/sidecar/ovsp4rt.cc
```diff
@@ -375,8 +381,8 @@ void PrepareFdbTableEntryforV4VxlanTunnel(
           param->set_param_id(GetParamId(
               p4info, L2_FWD_TX_TABLE_ACTION_POP_VLAN_SET_VXLAN_UNDERLAY_V6,
               ACTION_PARAM_TUNNEL_ID));
-          // TODO(derek): 8-bit vni (es2k, ipv6)
-          param->set_value(EncodeByteValue(1, learn_info.tnl_info.vni));
+          // tunnel_id is bit<20>
```
5abeel, line 384: Is this comment correct? IPv6 uses 20-bit VNI instead of 24-bit? Please double-check

ovs-p4rt/sidecar/ovsp4rt.cc
ffoulkes: Because the field it's being encoded for is named `tunnel_id`, not `vni`, and the value is 20 bits, not 24.

Each use of the `vni` input field has its own semantics, which are related to the field in the P4Runtime request to which the value is written.

Using a purpose-specific function allows me to encapsulate the semantics (the high byte is ANDed with 0x0F instead of 0xFF). The function name alerts the reader to the different semantics.

I'm being consistent by using `EncodeTunnelId` for all 20-bit tunnel ID values and `EncodeVniValue` for all 24-bit VNI values, and I'm avoiding false equivalence by not using the same encoding function in both cases.

ovs-p4rt/include/ovsp4rt/ovs-p4rt.h
ffoulkes: I haven't submitted the OVS PR yet. Thanks to one or more recent software updates, WSL's connection to the outside network goes dead and I have to reboot to recover. This leads to delays -- sometimes long ones, because it's late in the day or I'm frustrated and need to take a break -- and I frequently forget, when I return to my computer, that I have an unpushed PR to return to. That's what happened in this case. And when I found it this morning, and went to push it ... you guessed. The network connection was dead again. So I broke for lunch.

ovs-p4rt/sidecar/ovsp4rt.cc
ffoulkes: I'm positive.

```json
actions {
  preamble {
    id: 23849990
    name: "linux_networking_control.pop_vlan_set_vxlan_underlay_v6"
    alias: "pop_vlan_set_vxlan_underlay_v6"
  }
  params {
    id: 1
    name: "tunnel_id"
    bitwidth: 20
  }
}
```

ffoulkes: I mentioned this in the PR comment:

> Implemented a new `EncodeTunnelId()` function to use in place of `EncodeVniValue()` when the destination is a tunnel_id field. It encodes the low-order 20 bits of a uint32_t parameter. _Note that these VNI values are 20 bits, not 24._

ffoulkes: Stuff like this is why I'm increasingly concerned about the lack of a test suite.

ovs-p4rt/include/ovsp4rt/ovs-p4rt.h
ffoulkes: > What are the repercussions of changing this to 32-bit here and not in OVS codebase?

Not an issue if OVS is built `--with-ovsp4rt`. OvS uses the copy in the networking-recipe install tree.

Bad if it's built with `--with-p4ovs`. OvS uses its private copy.

ffoulkes:
I wrote a super-minimal unit test with a single test case for `PrepareTunnelTermTableEntry()`, which is one of the other two functions that I've modified to use `EncodeTunnelId()`. I discovered that it uses the `vni` input _twice_ -- once as a bit<24> in a match key, and once as a bit<20> when setting the `tunnel_id`.

Yikes.

ffoulkes:
Unit test results:
```text
dfoster@dgfoster-mobl1:~/work/latest$ (cd build; ctest -L ovsp4rt)
Test project /home/dfoster/work/latest/build
      Start  6: encode_tunnel_id_test
 1/12 Test  #6: encode_tunnel_id_test ..............   Passed    0.23 sec
      Start  7: encode_host_port_value_test
 2/12 Test  #7: encode_host_port_value_test ........   Passed    0.02 sec
      Start  8: geneve_encap_v4_table_entry_test
 3/12 Test  #8: geneve_encap_v4_table_entry_test ...   Passed    0.05 sec
      Start  9: geneve_encap_v6_table_entry_test
 4/12 Test  #9: geneve_encap_v6_table_entry_test ...   Passed    0.03 sec
      Start 10: geneve_encap_v4_vlan_pop_test
 5/12 Test #10: geneve_encap_v4_vlan_pop_test ......   Passed    0.03 sec
      Start 11: prepare_l2_to_tunnel_test
 6/12 Test #11: prepare_l2_to_tunnel_test ..........   Passed    0.03 sec
      Start 12: tunnel_term_v4_table_test
 7/12 Test #12: tunnel_term_v4_table_test ..........   Passed    0.04 sec
      Start 13: tunnel_term_v6_table_test
 8/12 Test #13: tunnel_term_v6_table_test ..........   Passed    0.04 sec
      Start 14: vxlan_encap_v4_table_entry_test
 9/12 Test #14: vxlan_encap_v4_table_entry_test ....   Passed    0.03 sec
      Start 15: vxlan_encap_v6_table_entry_test
10/12 Test #15: vxlan_encap_v6_table_entry_test ....   Passed    0.03 sec
      Start 16: vxlan_encap_v4_vlan_pop_test
11/12 Test #16: vxlan_encap_v4_vlan_pop_test .......   Passed    0.04 sec
      Start 17: vxlan_encap_v6_vlan_pop_test
12/12 Test #17: vxlan_encap_v6_vlan_pop_test .......   Passed    0.04 sec

100% tests passed, 0 tests failed out of 12

Label Time Summary:
ovsp4rt    =   0.61 sec*proc (12 tests)

Total Test time (real) =   0.64 sec
```

PR #586, opened 2024-07-27
[ovsp4rt] Refine preprocessor conditionals

ffoulkes:
- In #if defined(<target1>) / #else combinations, changed the #else to #elif defined(<target2>).

- In #if defined() / #elif defined() combinations, made sure there was an #else / #error stanza before the #endif.

- Modified #error messages to use the same text as krnlmon.

- Added or removed vertical white space around some preprocessor conditionals.

- Added some #endif end-of-line comments.


5abeel (approved):
LGTM

PR #588, opened 2024-07-31
[ovsp4rt] Export list of unit tests to parent scope

ffoulkes:
- Reworked the process of accumulating the list of unit tests to be built by the `ovsp4rt-unit-tests` target so it works across subdirectories.


5abeel (approved):
LGTM

PR #594, opened 2024-08-03
[ovsp4rt] Split prepare_l2_to_tunnel_test into two tests

ffoulkes:
- Divided `prepare_l2_to_tunnel_test` into separate IPv4 and IPv6 tests (one for each function being tested).

- Defined separate insert_entry and remove_entry test cases for each test.

- Updated both tests to conform to current practices.

- Deleted the original test.


5abeel (approved):
lgtm

PR #599, opened 2024-08-05
[ovsp4rt] Implement tunnel_term_v4_table_test

ffoulkes:
- Unit test for PrepareTunnelTermTableEntry().

> [!NOTE]
> `PrepareTunnelTermTableEntry()` adds a tunnel_id match field to the table entry when it builds the request. _This table does not have a tunnel_id match field._ What gives?


5abeel (approved):
LGTM

PR #604, opened 2024-08-07
[ovsp4rt] Implement vlan_pop_table_test

ffoulkes:
- Unit test for PrepareVlanPopTableEntry().

> [!NOTE]
> `vlan_id`, which is nominally bit<12>, is encoded as bit<8>. What gives?


5abeel (approved):
lgtm

PR #608, opened 2024-08-07
[p4runtime] Update requirements for python wheel

ffoulkes:
Requirements:
- grpcio >= 1.65.2 (was 1.59.2)
- googleapis-common-protos >= 1.63.0 (was 1.61.0)

Updated wheel version to 2024.8.0


5abeel (approved):
LGTM

5abeel:
I think we also need to update the python wheel spec in the p4runtime-dev. Current spec there shows grpc==1.17.2

https://github.com/ipdk-io/p4runtime-dev/blob/07f671755e32bed4edc8fab32c86c4d2dedb607c/py/setup.cfg#L24-L25

ffoulkes:
> I think we also need to update the python wheel spec in the p4runtime-dev. Current spec there shows grpc==1.17.2

I'd like to leave the spec in p4runtime-dev as-is for the time being. We currently build the python wheel using our own configuration files in networking-recipe.

PR #611, opened 2024-08-08
[ovsp4rt] Implement EncodeTunnelId() function

ffoulkes:
- Introduced an `EncodeTunnelId()` function, to be used instead of `EncodeVniValue()` when encoding a `vni` input as the value of a `tunnel_id` action parameter. The two functions are currently identical.

Note that each call site specifies the value for an ACTION_PARAM_TUNNEL_ID.

Extracted from https://github.com/ipdk-io/networking-recipe/pull/585.

----

> [!NOTE]
> The first step is to introduce the new function and run the unit test suite to check for errors, which I have done. The test suite explicitly checks all occurrences of the `tunnel_id` action parameter.
>
> The next step requires that we widen the `vni` input field to 32 bits.

### DIFF
--- a/ovs-p4rt/sidecar/ovsp4rt.cc
+++ b/ovs-p4rt/sidecar/ovsp4rt.cc
@@ -48,6 +48,14 @@ std::string EncodeByteValue(int arg_count...) {
   return byte_value;
 }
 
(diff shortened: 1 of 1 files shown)


5abeel (approved):
LGTM

PR #612, opened 2024-08-08
[ovsp4rt] Revise geneve and vxlan encap tests

ffoulkes:
1. Created a unique subclass for each encapsulation test, making it possible for the test to have multiple test cases.
2. Extracted the existing test code into a `CheckAction()` method in the test class. This makes the test bodies smaller and much easier to understand.
3. Defined a `CheckMatches()` method to verify the match fields. For these tests, they all match on the VNI.
4. Created `remove_entry` test cases to complement the existing `insert_entry` cases.

----

I made all the changes at the same time because I need the tests to verify the 24-bit VNI changes. This results in a rather unwieldy PR. I can break it into smaller PRs, if you would like.

The changes are actually simpler than they look. The problem is that the Extract Method refactoring resulted in a substantial portion of the original code being indented. The only difference is in the leading white space, but GitHub sees this as a large insertion and a large deletion occurring in the same area of the file. Pfui.

### DIFF
--- a/ovs-p4rt/sidecar/testing/geneve_encap_v4_table_entry_test.cc
+++ b/ovs-p4rt/sidecar/testing/geneve_encap_v4_table_entry_test.cc
@@ -11,72 +11,97 @@
 
 namespace ovsp4rt {
 
+constexpr bool INSERT_ENTRY = true;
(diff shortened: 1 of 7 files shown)


5abeel (approved):
LGTM

PR #613, opened 2024-08-09
[ovsp4rt] Update tests to use BaseTableTest class

ffoulkes:
- Modified several tests to use BaseTableTest as their parent class.

The immediate motivation is to eliminate local definitions of `DecodeVsiValue()`, in preparation for supporting 24-bit VSIs.


5abeel (approved):
LGTM

PR #633, opened 2024-08-13
[ovsp4rt] Base class for tunnel tests

ffoulkes:
- Replacement for `Ipv4TunnelTest` and `Ipv6TunnelTest`.
- Based on `BaseTableTest` instead of `TableEntryTest`.


5abeel (approved):
LGTM

PR #643, opened 2024-08-14
[ovsp4rt] Update unit tests for wide VNIs

ffoulkes:
- Widened a temporary variable from 16 to 32 bits.
- Added a couple of 24-bit VNI test cases.

Extracted from PR https://github.com/ipdk-io/networking-recipe/pull/616 so the test updates can be merged separately.

> [!NOTE]
> These are safe changes. They augment what they unit tests support. They have no effect on encodings or the public API.


5abeel (approved):
lgtm

PR #674, opened 2024-08-26
[ovsp4rt] Implement Client class (work in progress)

ffoulkes:
This is a project to improve the testability of ovsp4rt. See issue https://github.com/ipdk-io/networking-recipe/issues/701 for details.

- Modified ovsp4rt.cc to use a `Client` object instead of an `OvsP4rtSession` object to communicate with the P4Runtime server.

- Made each public C API function a wrapper around a C++ function that accepts a Client object as a parameter.  This allows a unit test to replace the object with a test double (e.g. a mock).

- Moved the C API functions to another file, separating the user interface from the implementation. This allows a substitute front end to be used for testing.

- Implemented `HaveEntry` predicates as wrappers around the `GetEntry` functions. Modified the code to use the predicates when the purpose of the read request is to test for the existence of an entry.

> [!NOTE]
> The `BUILD_CLIENT` cmake option is enabled by default in this branch. We cannot compile the modified version of `ovsp4rt.cc` without it.


ffoulkes:
Superseded by PR https://github.com/ipdk-io/networking-recipe/pull/727.
Closing without deleting, so the original PR is available for reference.

PR #693, opened 2024-10-02
[ovsp4rt] Improve journal encoding

ffoulkes:
- Encode and validate IPv6 addresses as arrays of uint16_t. This allows us to provide better feedbv
- Make a couple of minor changes to encode_addr_test.


5abeel (approved):
LGTM

PR #696, opened 2024-10-12
Update CMakePresets.json file

ffoulkes:
- Enabled CMAKE_EXPORT_COMPILE_COMMANDS. This option instructs CMake to generate a `compile_commands.json` file in the build directory, for use by other tools (such as `clang-tidy`)

- Expanded CMAKE_INSTALL_PREFIX to an absolute path, which is what happens when you set `CMAKE_INSTALL_PREFIX` on the CMake command line.


5abeel (approved):
LGTM

PR #697, opened 2024-10-14
Define list of values for P4OVS_MODE variable

ffoulkes:
- Set the STRINGS property on the P4OVS_MODE variable to a list of supported values. The user may choose from this list in `ccmake` or the cmake GUI.


5abeel (approved):
LGTM

PR #698, opened 2024-10-24
Fix problem setting property on P4OVS_MODE

ffoulkes:
- Moved the set_property() for the STRINGS property on P4OVS_MODE after its definition. This fixes a CMake configuration error that cropped up in the Tofino build.


5abeel (approved):
LGTM

PR #704, opened 2024-11-01
Minor improvements to README file

ffoulkes:
- Sanitized a couple of file paths.

- Changed "pre-load" to "preload". The CMake documentation uses the hyphenated form, but I find it distracting, especially when it's juxtaposed with "preset".


5abeel (approved):
LGTM

PR #707, opened 2024-11-03
Mark development options as advanced

ffoulkes:
- Marked the BUILD_JOURNAL and BUILD_SPIES variables as advanced.
  They are internal options, used to enable features that are
  still under development, and should normally be hidden.


5abeel (approved):
LGTM

PR #714, opened 2024-12-10
[ovsp4rt] Minor maintenance

ffoulkes:
- Renamed the Journal "schema" element to "version".

- Added `const` qualifier to a couple of parameters.

> [!NOTE]
> This is a passive update. The Journal code is not currently built unless `BUILD_JOURNAL` is enabled.


5abeel (approved):
LGTM

Issue #701, opened 2024-10-28
[ovsp4rt] Public API functions are difficult to test


ffoulkes:
### Problem statement

The functions in ovsp4rt's public API are inherently difficult to unit test. This is because each function embeds the logic to establish a client connection to the P4Runtime server.

```c
  // Start a new client session.
  auto status_or_session = ovsp4rt::OvsP4rtSession::Create(
      grpc_addr, GenerateClientCredentials(), absl::GetFlag(FLAGS_device_id),
      absl::GetFlag(FLAGS_role_name));
  if (!status_or_session.ok()) return;

  // Unwrap the session from the StatusOr object.
  std::unique_ptr<ovsp4rt::OvsP4rtSession> session =
      std::move(status_or_session).value();
  ::p4::config::v1::P4Info p4info;
  ::absl::Status status =
      ovsp4rt::GetForwardingPipelineConfig(session.get(), &p4info);
  if (!status.ok()) return;
```

In addition, all communication with the server is done via function calls that couple the code tightly to the client session.

This makes it extremely difficult to isolate the body of the function so it can be tested independently of the P4Runtime server.

### Proposed solution

The class is solution to this kind of problem is to introduce an object that provides an abstract interface to the server, and then pass that object to a function containing the core logic, a technique known as _dependency injection_. We can then provide two (or more) implementations of the server interface: one that communicates with an actual P4Runtime server, and one (or more) that allow us to simulate the desired behavior of the P4Runtime server according to the needs of the individual test case.

To implement this, we introduce an ovsp4rt `Client` object (there's already a `Session` object) to handle the interface to the server, and we divide each public API function into two parts. One instantiates a Client object and passes it to the function that implements the core logic:

```c
void ovsp4rt_config_fdb_entry(struct mac_learning_info learn_info,
                              bool insert_entry, const char* grpc_addr) {
  using namespace ovsp4rt;
  Client client;
  ConfigFdbEntry(client, learn_info, insert_entry, grpc_addr);
}
```

The other implements the core logic:

```c
void ConfigFdbEntry(Client& client, struct mac_learning_info learn_info,
                    bool insert_entry, const char* grpc_addr) {
  absl::Status status;

  // Start a new client session.
  status = client.connect(grpc_addr);
  if (!status.ok()) return;

  // Fetch P4Info object from server.
  ::p4::config::v1::P4Info p4info;
  status = client.getPipelineConfig(&p4info);
  if (!status.ok()) return;

  .
  .
}
[...]

PR #716, opened 2024-12-16
[ovsp4rt] Remove session() method from Client class

ffoulkes:
The purpose of the `Client` class is to replace the discrete logic that interacts with the P4Runtime server with an object that implements an abstract interface. It should not expose the session object to the user.

### DIFF
--- a/ovs-p4rt/sidecar/client/ovsp4rt_client.h
+++ b/ovs-p4rt/sidecar/client/ovsp4rt_client.h
@@ -24,9 +24,6 @@ class Client : public ClientInterface {
   // Connects to the P4Runtime server.
   virtual absl::Status connect(const char* grpc_addr);
 
-  // Returns a pointer to the ovsp4rt session object.
-  virtual OvsP4rtSession* session() const { return session_.get(); }
-
   // Gets the pipeline configuration from the P4Runtime server.
   virtual absl::Status getPipelineConfig(::p4::config::v1::P4Info* p4info);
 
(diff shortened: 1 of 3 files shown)


ovs-p4rt/sidecar/client/ovsp4rt_client.h
```diff
@@ -24,9 +24,6 @@ class Client : public ClientInterface {
   // Connects to the P4Runtime server.
   virtual absl::Status connect(const char* grpc_addr);
 
-  // Returns a pointer to the ovsp4rt session object.
-  virtual OvsP4rtSession* session() const { return session_.get(); }
```
5abeel, line 28: Isnt this needed for future use?

ffoulkes: I thought so at first, but I've changed my mind.

The purpose of the `Client` class is to replace the discrete logic that interacts with the P4Runtime server with an object that implements an abstract interface.

- `ClientInterface` is an abstract class that defines the interface.
- `Client` is an implementation of `ClientInterface` that replaces the current discrete logic.
- `ClientMock` is a mock implementation of `ClientInterface`.

See issue https://github.com/ipdk-io/networking-recipe/issues/701 for more information.

5abeel (approved):
LGTM

ovs-p4rt/sidecar/client/ovsp4rt_client.h
5abeel: Ack

PR #719, opened 2024-12-26
[ovsp4rt] Implement journaling client

ffoulkes:
- Created `JournalClient`, a subclass of `Client` that is instrumented to record interactions between OVS and the P4Runtime server.

> [!NOTE]
> Extracted from PR #674. Currently excluded from the build.


5abeel (approved):
LGTM

PR #720, opened 2024-12-26
[ovsp4rt] Separate C API functions from rest of code

ffoulkes:
- Moved the C API wrapper functions to a separate source file. This makes it possible to substitute a different implementation of the API, for testing.

Extracted from PR #674. Works with an updated version of ovsp4rt.cc that has not been committed. Currently excluded from the build.


5abeel (approved):
LGTM

PR #721, opened 2024-12-26
[ovsp4rt] Implement journaling API

ffoulkes:
- Created a variant of the public API that uses JournalClient objects.

Extracted from PR #674. Works with several files that have not yet been committed, or that are being committed separately. Currently excluded from the build.


5abeel (approved):
LGTM

Issue #618, opened 2024-08-10
[ovsp4rt] Public APIs do not return status


ffoulkes:
The functions that OVSP4RT exports for use by OVS should return error status when they are unable to complete the P4 update(s) successfully.

PR #734, opened 2025-01-09
[ovsp4rt] Update C++ API functions to return absl::Status

ffoulkes:
- Modified the functions in the internal C++ API to return `absl::Status`, allowing test programs to verify whether the function encountered an error.
- Modified the C API wrappers to ignore the returned Status explicitly. The public APIs currently have a `void` return type.

These changes only apply when BUILD_CLIENT is enabled.

> [!NOTE]
>   Issue [#618](https://github.com/ipdk-io/networking-recipe/issues/618) proposes that the public APIs should return status. The Status returned by the core APIs will give them the means to do so, should we decide to implement this change.

### DIFF
--- a/ovs-p4rt/sidecar/newovsp4rt.cc
+++ b/ovs-p4rt/sidecar/newovsp4rt.cc
@@ -1,6 +1,19 @@
 // Copyright 2022-2025 Intel Corporation
 // SPDX-License-Identifier: Apache-2.0
 
+// Revision of ovsp4rt.cc to improve testability.
+// *** UNDER CONSTRUCTION ***
+//
+// - Replaced the OvsP4rtSession object with a Client object,
+//   which provides an abstract interface to the P4Runtime
+//   server.
+//
+// - Split each ovsp4rt API into an internal C++ function
+//   that implements the core logic of the API, and a
+//   C wrapper function.
+//
+// - Moved the C wrapper functions to a separate file.
+
 #include <arpa/inet.h>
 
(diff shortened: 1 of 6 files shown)


ovs-p4rt/sidecar/newovsp4rt.cc
```diff
@@ -2269,39 +2269,29 @@ void DoConfigFdbEntry(ClientInterface& client,
...
       }
     }
 
-    status =
-        ConfigFdbTunnelTableEntry(client, learn_info, p4info, insert_entry);
-    if (!status.ok()) {
-      // Ignore errors (why?)
-    }
+    // Ignore errors (why?)
+    (void)ConfigFdbTunnelTableEntry(client, learn_info, p4info, insert_entry);
```
5abeel, line 2277: Is the `(void)` addition to suppress compiler warnings?

5abeel (approved):
LGTM

ovs-p4rt/sidecar/newovsp4rt.cc
ffoulkes: Yes. It is also a way of letting the reader know that we _know_ the function returns a value and are choosing to ignore it.

The _comment_ is shorthand for "I really wish the original developer had either handled the returned the return or left a comment explaining why they chose to ignore it. I really wish they had done a better job, but I'm not in a position to do anything about it right now. I will, however, leave a marker in the code for the benefit of future developers."

PR #735, opened 2025-01-09
[ovsp4rt] Extract ovsp4rt_str_to_tunnel_type()

ffoulkes:
- Moved ovsp4rt_str_to_tunnel_type() to a separate source file for C api functions that are not wrappers.

This change only applies when BUILD_CLIENT is enabled.


5abeel (approved):
LGTM

PR #737, opened 2025-01-12
[ovsp4rt] Implement GetTxAccVsiPort()

ffoulkes:
- Extracted code from `ConfigFdbEntry()` and `ConfigSrcPortEntry()` to a new `GetTxAccVsiPort()` function. This reduces the complexity of the calling functions, which don't need to know how the port is obtained, and it removes duplicate code.

See later comments in the PR for AI analysis of this refactoring.

This PR is ES2K-specific. `GetTxiAccVsiPort` is only called by ES2K functions.

> [!NOTE]
> The purpose of this PR is to *refactor*. I propose to make the mutability changes noted in the comments in a second PR, and any other changes in subsequent PRs.


ffoulkes:
Here is ChatGPT's assessment of the first refactoring:

Let's compare the original and refactored code and identify key differences to determine if they are functionally equivalent:

### Key Changes:

1. **Refactoring of TxAccVsi Port Logic:**

   - **Original Code:** The logic to obtain the src_port was handled directly within DoConfigFdbEntry. It involves calling GetTxAccVsiTableEntry, processing the ReadResponse, and extracting the src_port.
   - **Refactored Code:** This logic has been extracted into a new function, GetTxAccVsiPort, which encapsulates the same functionality. The refactored DoConfigFdbEntry then calls this function and updates learn_info.src_port accordingly.

2. **GetTxAccVsiPort:**

   In the refactored version, the code that fetches the src_port from the TxAccVsiTableEntry and processes the response has been moved to a separate function (GetTxAccVsiPort). This makes the DoConfigFdbEntry function cleaner and easier to follow. The GetTxAccVsiPort function itself returns the processed src_port, which is then used in the DoConfigFdbEntry function.

3. **`host_sp` Handling:**

   In the original code, the host_sp variable is calculated inline within the DoConfigFdbEntry function.
In the refactored code, host_sp is calculated in the GetTxAccVsiPort function and then returned to DoConfigFdbEntry.

4. **Error Handling:**

   Both versions handle errors similarly, but the error handling is encapsulated in the `GetTxAccVsiPort` function in the refactored version. It checks if the result is an error (`!status.ok()`) before continuing.

### Functional Equivalence:

- The core logic of **connecting to the client, fetching P4Info, checking tunnel entries**, and **configuring the FDB entries** remains unchanged in both versions.
- The **refactoring** does not alter the behavior of the system. It simply moves the logic of obtaining the `src_port` into a separate function (`GetTxAccVsiPort`), which is then invoked in `DoConfigFdbEntry`. This refactor improves the modularity and readability of the code without changing its overall functionality.
- The handling of the `learn_info.src_port` and the update logic is preserved in the refactored code, so the flow of the program remains the same.

### Conclusion:

The original and refactored code are **functionally equivalent**. The refactoring only reorganizes the code to improve clarity and modularity, without altering the core behavior.

ovs-p4rt/sidecar/newovsp4rt.cc
```diff
@@ -2110,6 +2110,49 @@ absl::StatusOr<::p4::v1::ReadResponse> GetTxAccVsiTableEntry(
...
+      if (param_id == param.param_id()) {
+        const std::string& s1 = param.value();
+        // TODO(derek): why make a mutable copy?
+        std::string s2 = s1;
+        for (int param_bytes = 0; param_bytes < 4; param_bytes++) {
+          host_sp = host_sp << 8 | int(s2[param_bytes]);
+        }
+        // TODO(derek): This break exits to the outermost for loop,
+        // advancing to the next table entry. Is this correct?
+        // Would `return host_sp` be better?
```
ffoulkes, line 2148: If we replace `break` with `return host_sp`, we can (1) define and initialize `host_sp` within the `if` statement, and (2) change line 2153 to `return 0`.

PR #754, opened 2025-03-04
Es2k workflow check

5abeel:
Adding workflow to build and test against `es2k` target.

Two self-hosted runners have been added in lab running Ubuntu-22.04 and will be used to trigger jobs using the `es2k-runner` label


n-sandeep (approved):
LGTM

PR #763, opened 2025-03-25
Update Stratum submodule to fix TLS cert generation

5abeel:
Update stratum submodule to pick up the TLS certification fix


vsureshkumarp (approved):
LGTM..:)

PR #765, opened 2025-03-25
Workflow build check name change

5abeel:
The workflow is to check build only and unit tests arent performed (atleast not yet). So, changing name from `es2k_build_and_test` to `es2k_build_check`


n-sandeep (approved):
LGTM

PR #771, opened 2025-04-01
Adding version.json to set project version number

5abeel:
Instead of setting a project version number in CMakeLists.txt, setting the value in a version.json file. The file is then parsed to set the project version number


weiqiand (approved):
LGTM

